PR #13, opened 2016-12-08
Write Output of AMW Run Script to Logfile

phil-pona:
Implemented first Version. @yvespp could you please review and Test on different AS and deployments?

### DIFF
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/RunSystemCallService.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/RunSystemCallService.java
@@ -62,6 +62,7 @@ public class RunSystemCallService
 	 * @param folder
 	 *               - the relative path of the configuration-folder for
 	 *               which the execution scripts shall be invoked.
+	 * @return Result Output of Script
 	 * @throws ScriptExecutionException
 	 *                - if the execution of one of the scripts was not
 	 *                successful. If the script was executed but had errors
@@ -69,15 +70,18 @@ public class RunSystemCallService
 	 *                exception should contain the given information provided
 	 *                by the process itself.
 	 */
-	public void getAndExecuteScriptFromGeneratedConfig(String folder)
+	public String getAndExecuteScriptFromGeneratedConfig(String folder)
 			throws ScriptExecutionException {
 
 		List<File> scriptPath = getScriptFiles(folder);
 
+		StringBuilder result = new StringBuilder();
+
 		for (File file : scriptPath) {
 			String filePath = makeScriptExecutableAndGetAbsolutePath(file);
 			if (filePath != null) {
-				executeScript(filePath);
+				result.append("output of Script " + filePath);
+				result.append(executeScript(filePath));
 			} else {
 				String message = "File permissions of "
 						+ file.getName()
@@ -87,6 +91,7 @@ public class RunSystemCallService
 						REASON.PERMISSION);
 			}
 		}
+		return result.toString();
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/RunSystemCallService.java
```diff
@@ -50,6 +50,8 @@
 public class RunSystemCallService
 {
 
+	private String lineSeparator = System.getProperty("line.separator");
```
yvespp: System.lineSeparator()

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/RunSystemCallService.java
```diff
@@ -136,6 +147,7 @@ private void runSystemCall(String command)
 			InputStreamReader inread = new InputStreamReader(buf);
 			BufferedReader bufferedreader = new BufferedReader(inread);
 			StringBuilder sb = new StringBuilder();
+			StringBuilder systemCallResult = new StringBuilder();
 			String line;
 			while ((line = bufferedreader.readLine()) != null) {
 				// TODO schreibe resultat in eine logdatei
```
yvespp: The TODOs can be removed now

phil-pona:
very good point with the result as String, instead of passing a stream, I pass the path to the logfile. The interface seams smaller to me that way. I'm going to test the new integration first on my side.

phil-pona:
@yvespp please check again and test on deployments

yvespp:
Here is what I found in my tests:
* I removed all manual redirects in the rundscript like this:
  * old: `echo "start script" >> ${deployment.amwLogFile}`
  * new: `echo "start script"`
* The output shows up in the Log View, however the log of the predeploy disappears as soon as the deploy step starts. I guess this is because the file gets overwritten when den deploy starts. Can you instead append to the logo file if it already exists?
* I put `echo "@{test}"` in the deploy script which should show up in the deployment details but it doesn't.

@phil-pona can you fix the two points?

yvespp:
Also stuff that gets logged to stderr will not show up in the log.
To get the error stream you can call `p.getErrorStream()`.

In ProcessBuilder there is an option to redirect stderr to stdout so you only have to deal with one stream: http://docs.oracle.com/javase/6/docs/api/java/lang/ProcessBuilder.html#redirectErrorStream(boolean)

phil-pona:
- changed to append mode
- redirected ErrorStream to Logoutput
- values with @{test} are only displayed in the deployment view if an error occurs, this behavior did not change

yvespp:
> values with @{test} are only displayed in the deployment view if an error occurs, this behavior did not change

Didn't know that, that's then.

@phil-pona I tested again, here is what I found:
* The Output gets redirected to a file with name `156704_node_01_script_output.log` but the ${deployment.amwLogFile} points to 156704_node_01.log. We write some output directly to the log using this variable which now ends up in the wrong file. This is needed because the predeploy script is run in the background. In the run script: `python -m refact.amwDeployRunner >> ${deployment.amwLogFile} 2>&1 &`
Can you make `deployment.amwLogFile` point to the actual log file?
* I don't see the annotated logs (e.g. `@{test}`) when the deployment fails. I tested this by adding an exit 1 at the end of the scripts. Maybe related to the first problem?
* Can you remove the brackets ouround this log: `Executing Script: (/data/amw/build/amw/Local_node_01_156707_predeploy/amwScript/AmwRunScript.sh)`
Looks a bit weird to me :-)

Such a small change, so much work 😃

phil-pona:
fixed the two things, we need to test it well because of concurrency issues.

lets discuss the ${test} thing face to face next time

phil-pona:
ok can be merged.

**Feature Breaking Change**
This pullrequest changes the behavior of the Deployment Logfile, since the java process now writes the script output directly into the same logfile. 
The output via ``${deployment.amwLogFile}`` in the deployment script appears first in the logfile.
The output created by the script with eg. (echo "test") will be logged after the other logoutput.

Workaround: choose a second Logfile to write asynch output to it. 
```
echo "test" >> ${deployment.targetLogPrefix}_secondlog.log
```

Issue #32, opened 2016-12-27
Initial model doesn't work with h2


yvespp:
I tried to generate a new h2 database using the new data set from #22. I want to use this as base for #11.
It fails because the sqls contains `to_date` and `to_timestamp`  (maybe more?) which h2 doesn't know.

Ideas:
* Rewrite the affected queries in Liquibase syntax
* Define a function for h2: http://dateconvertion.blogspot.ch/2011/02/date-conversion-from-oracle-format-to.html

PR #37, opened 2017-01-16
Make initial db script usable for h2

yvespp:
Implements #32.
Includes #36 -> I will rebase after merge.

I used jdbc escaping for dates and timestamps, which is supposed to work with oracle, postgresql and h2.
I tested h2 only.

@phil-pona: do you have time to test with postresql and oracle?


AMW_business/src/main/resources/liquibase/changesets/base/data/exampleapp/app.data.xml
```diff
@@ -16,32 +16,29 @@
 		</sql>
 	</changeSet>
 
-	<changeSet author="initial-data" id="example-resources-pg" dbms="postgresql">
+	<changeSet author="initial-data" id="example-resources">
```
phil-pona: not sure about that, I would rather define the supported dbms explicitly than the "hack" with the precondition

AMW_business/src/main/resources/liquibase/changesets/base/data/initial.data.xml
```diff
@@ -18,25 +18,16 @@
         </sql>
     </changeSet>
 
-    <changeSet author="initial-data" id="initial-data-base-release-1-pg" dbms="postgresql">
```
phil-pona: is not going to work, the boolean for MAINRELEASE needs to be an integer for oracle

PR #43, opened 2017-02-02
Makes AppVersions in Deployment Rest call optional

phil-pona:
For some UseCases the AppVersion Parameter on the Deployment Rest API ist optional. 

@Yvesp could you please verify?


yvespp:
@phil-pona when I deploy via rest without ApplicationWithVersion the deplyoment is show without apps in the deployment view, even if the appserver has apps...

Maybe ApplicationWithVersion can only be omitted if the appserver has no apps?
Or what's the usecase for this?

PR #47, opened 2017-02-15
defined "is" as default comparator for tracking id filter

### LINKED ISSUE #45: Tracking Id deployment Filter: set default to is
The default comparison operator for the Tracking Id deployment filter is currently `<` which is used rarely.

The default should be set to `is`,

RetGal:
This closes #45


yvespp:
Changed it so it applies to all filters of IntegerType

Issue #64, opened 2017-03-06
Go to Resource Type with NullFilterViewHandler breaks editResourceView


yvespp:
With the filter from #41 enabled "Go to" -> "Resource type" causes the editResourceView to break:
* relations can't be selected
* unresolved dependencies are show, they should only be show on instances not types.
* on none default resource type (e.g. Webservice) clicking on a unresolved relation causes a NullPointer : /resources/mobi/editRelatedProperties.xhtml @82,138 listener="#{relationDataProvider.loadResourceGroupsForType(key, unresolved)}": javax.ejb.EJBException: java.lang.NullPointerException

If I open the screen via "resourceList" -> "Edit Resource Type" then everything works correctly.

Same issue I found on Wildfly 10:  https://github.com/liimaorg/liima/issues/9#issuecomment-284244448

On our installation I disabled the NullFilterViewHandler by removing it in AMW_web/src/main/webapp/WEB-INF/faces-config.xml.

yvespp:
Fixed by #66
Should be addressed in #9

PR #66, opened 2017-03-08
disable NullFilterViewHandler as it causes problems on editResourceView

yvespp:
See #64

We should merge this asap as master is broken atm. 
I just disabled the filter so it can be enabled for testing.


yvespp:
@phil-pona should we also make a new release?

phil-pona:
yes I guess we should, since you're going to deploy it on your prod, don't you?

yvespp:
I disabled it manually and is already on prod.

Am 08.03.2017 11:29 schrieb "Thomas Philipona" <notifications@github.com>:

PR #70, opened 2017-04-03
Don't lock on nodejob update, backoff if lock fails

yvespp:
Reduces the locks of a deployment dramatically and also reduces the load on the DB when locks happen.
Fixes a bug where locking constantly failed with deployments with more than 8 nodes.
We also had a problem with locks, when the DB was slow.

### DIFF
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/LockingService.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/LockingService.java
@@ -58,15 +58,17 @@ public class LockingService
 	 */
 	public DeploymentEntity lockDeployment(Integer id) {
 		int count = 0;
+		int backoff = 300;
 		
-		log.info("Locking deployment " + id);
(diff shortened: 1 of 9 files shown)


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/generator/control/LockingService.java
```diff
@@ -58,15 +58,17 @@
...
 		for(; count <= lockRetries; count++) {
 			try {
 				return entityManager.find(DeploymentEntity.class, id, LockModeType.PESSIMISTIC_FORCE_INCREMENT);
 			} catch (LockTimeoutException e) {
-				log.info("Retrying to lock in 200 ms");
+				backoff = getBackoff(backoff);
+				log.info("Retring to lock deployment " + id + " in " + backoff + " ms");
 				try {
-					Thread.sleep(200);
+					Thread.sleep(backoff);
```
phil-pona: we should not use Thread.sleep in a Java EE Context, might look for a better solution here

Issue #87, opened 2017-05-31
Refactor CustomFilterComp and CustomFilter


yvespp:
In release v1.12.3 the deployment filter comparator for date was stuck in its default position. If you try to change it, it gets reset. This bug was introduced by the fix for Wildfly: https://github.com/liimaorg/liima/pull/63

If fixed it by partial reverting the pull: https://github.com/liimaorg/liima/compare/yvespp_deploymentcomparator_fix#diff-fc431175abc03858102a88ddfea21619

The CustomFilterComp and CustomFilter should be refactored:
* Move comperatorSelectionList selectedComperator props form CustomFilterComp to CustomFilter. CustomFilterComp should only hold the current filter list.
* After that https://github.com/liimaorg/liima/compare/yvespp_deploymentcomparator_fix#diff-603b0c275c3651fb4abb53c854c93bb3 can be changed to only iterate of filters and make it wildfly / eap 7 compatible.
* Move the construction of filters into a central place, example for currenct use of filters: [DeployScreenDataProvider ](https://github.com/liimaorg/liima/blob/yvespp_deploymentcomparator_fix/AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/deploy/DeployScreenDataProvider.java#L258), [DeploymentsRest ](https://github.com/liimaorg/liima/blob/yvespp_deploymentcomparator_fix/AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java#L446)
  * Only FilterType should be needed to construct a filter, all other values should be set automatically (valid comparators, default comparator, db column name etc.)

yvespp:
See bug #81

yvespp:
Fixed by #103

PR #101, opened 2017-06-19
Refactor custom filter comp and and custom filter

polandy:
refactored CustomFilter, solved issue  #87
- Moved comperatorSelectionList from  CustomFilterComp to CustomFilter.
- renamed the Boundary "DeploymentService" to "DeploymentBoundary"
- Using a builder with DeploymentFilterTypes as required argument to instantiate the CustomFilter.
- moved affected inner enums to seperate classes


polandy:
Should be merged into v.1.12.4_bugfix and not into master

PR #102, opened 2017-06-20
Bugfix resource list slow

### LINKED ISSUE #96: resourceList.xhtml slow with lots of resource instances
**Release**: v1.13.0-beta.1

Openen the list of resources on resourceList.xhtml takes a lot longer in 1.13 than it did in Liima 1.12:
* Selecting `APPLICATIONS` take on our Liima test environment 21 seconds, in prod with 1.12 it takes 1.5 seconds. There are about 600 apps.
* Selecting `Webservice` with about 440 resources takes also 21 seconds to load, in prod with 1.12 it takes 0.9 seconds.
* Resource list with less instances load a lot quicker.

RetGal:
Closes #96 
- resourceList selecting `APPLICATION` dropped from ~25s to ~4s
- resourceList selecting `Webservice` dropped from ~12s to <3s
- `Add shakedown test` dropped from ~13s to <1.5s

### DIFF
--- a/AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/security/SecurityDataProvider.java
+++ b/AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/security/SecurityDataProvider.java
@@ -56,6 +56,10 @@ public class SecurityDataProvider implements Serializable{
 
 	private Map<String, Boolean> canDeleteResource = new HashMap<>();
 
+	private Boolean canAddShakedownTest;
+
+	private Boolean canExecuteShakeTestOrder;
+
 	/**
 	 * Use {@link PermissionBoundary#hasPermission(Permission)} instead
 	 * @param permissionValue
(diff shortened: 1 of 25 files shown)


RetGal:
This pull includes Refactor custom filter comp and and custom filter (issue #87)

AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/security/SecurityDataProvider.java
```diff
@@ -46,6 +48,14 @@
...
+	private Map<String, Boolean> canEditResourceType = new HashMap<>();
+
+	private Map<String, Boolean> canDeleteResourceType = new HashMap<>();
+
+	private Map<String, Boolean> canEditResource = new HashMap<>();
+
+	private Map<String, Boolean> canDeleteResource = new HashMap<>();
+
 	/**
 	 * Use {@link PermissionBoundary#hasPermission(Permission)} instead
```
yvespp: Is this comment still valid? Is this method deprecated?

AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/security/SecurityDataProvider.java
```diff
@@ -46,13 +48,36 @@
 	@Inject
 	PermissionBoundary permissionBoundary;
 
+	private Map<String, Boolean> canEditResourceType = new HashMap<>();
```
yvespp: Maybe the maps should have cache in the name to make it clear?

yvespp:
The change for resourcesList adds a web session cache per user. How do we handle this when we switch to Angular? Via cache in the browser?
I'm asking because we could also move the cache into the ejbs and speed up all clients (Angular, Rest and JSF). But then we would need a mechanism to refresh the cache periodically (timer?).

Btw: is the old permission cache still there and JBoss has to be restart to refresh it?

@RetGal and @phil-pona what do you think?

RetGal:
Yes, the "old" permission cache is still in use (PermissionRepository), but  it's refreshed automatically every 20 minutes or if a permission is added or altered.

PR #104, opened 2017-06-21
Feature add roles

RetGal:
Includes bugfix_angular_ui pull as well


RetGal:
As I probably wont finish this today, a separate pull would be more appropriate

RetGal:
Just added the backend validation.

RetGal:
It's implemented the same way as for adding users: Simply type a role name which does not exist yet, as soon as you add a permission to it, the new role is created and the permission assigned.

PR #110, opened 2017-06-27
Feature editable identifier for resource relation entities

### LINKED ISSUE #78: Make identifier of resource relations editable
A user should be able to set the name of resource relation.
The identifier name should be shown as a property on the relation on the editResourceView, similar as the resource name is shown on a resource.

Example relations identifiers of batch:
image

RetGal:
Closes #78


yvespp:
@RetGal could you please make this pull request against master instead of release-1.12?

Issue #57, opened 2017-02-27
Rest: create resources


yvespp:
Like on the GUI, it should be possible to crate new Apps, AppServer and other resource instances via rest:

Use cases;
* Create a resource instance (or new release of a resource)
* Create relations between resources 
* Invoke copy from on new AppServer and App

yvespp:
Ref #51

RetGal:
Implemented 
- Create a resource instance (or new release of a resource)
- Create relations between resources

RetGal:
Implemented "copy from" in pull  #138

Issue #51, opened 2017-02-22
Resource relations REST - add relations between two ressources 


tbe1:
Extend the Rest API so that it is possible to add a relation between two resources. 
For example add a relation between a node and an appServer.

yvespp:
Implemented in #117

PR #117, opened 2017-07-04
Feature create resources

RetGal:
Implemented #51 and first part of #57:
- Create a resource instance (or new release of a resource)
- Create relations between resources


yvespp:
@RetGal I see you updated the javaee-api from 6.0 to 7.0. We still run on JBoss EAP 6.4 which only supports Java EE 6. We probably should wait till we are on JBoss EAP 7/Wildfly.

RetGal:
@yvespp Right, i replaced it by javaee-api 6.0-6 (6.0 caused `java.lang.ClassFormatError: Absent Code attribute in method that is not native or abstract in class file javax/ejb/EJBException` in tests)

PR #120, opened 2017-07-05
Distinguish old from new permissions

RetGal:
Distinguish old from new permissions in permissions dropdown (Fixes display issue 1.2)


RetGal:
Sorry, I was just some minutes too late

PR #114, opened 2017-07-03
Feature editable identifier for resource relation entities

RetGal:
As requested

PR #123, opened 2017-07-12
Feature editable identifier for resource relation entities

polandy:
Replaces Pull Request #114


yvespp:
There is still something wrong with renaming:
1. Add a relation (adIntern)
2. Add a second relation to the same resource (adIntern_1)
3. Try to rename RelationName adIntern_1 to adIntern. I get this warning: RelationName 'adIntern' is not allowed
4. Rename RelationName adIntern_1 to test.
All of the sudden there are three relations to adIntern...

yvespp:
I just noticed, that for the issue above it is only necessary to do this:
1. Add relation to resource (adIntern)
2. Rename RelationName from adIntern to test

RetGal:
Fixed

yvespp:
@RetGal I still get the same behavior im https://github.com/liimaorg/liima/pull/123#issuecomment-314788126 for some resource but not for all. adIntern has the bug, im doesn't...

Also I noticed that I can name two different resource relations the same if the are not from the same resource type.

RetGal:
@yvespp The effect of seemingly relation duplication after renaming occurs, if the slave resource exists in several releases. Currently only the relation in the selected release is being renamed, causing the observed effect:
1) before renaming: adIntern(adIntern_1) (in both, RL-14.10 and RL-16.10)
2) after renaming: adIntern(adIntern_1) (RL-14.10) and adIntern(test) (RL-16.10)

yvespp:
@RetGal I guess all releases need to be renamed in this case. Otherwise templates accessing such a relation via relationName are not valid depending on the release.
We also have to test if make a new release of the renamed relation creates problems.

RetGal:
@yvespp Fine, now it renames the relationName in all releases. The creation of a new release of a renamed relation went smoothly. 
The creation of duplicate relationNames of relations of different resource types is no longer possible.

PR #192, opened 2017-08-29
addRelation.xhtml: remove check if a resource is provided twice

yvespp:
With the check some resources with a lot of instances are still slow. Because provided resources is a rarely used feature it is enough to check only after the user has clicked the add the resource link.

Before:
before

After:
after


RetGal:
Looks reasonable. It was added in a separate commit.

PR #194, opened 2017-08-29
Corrected method call

### LINKED ISSUE #193: Copy from predecessor causes exception

RetGal:
Fixes #193


yvespp:
@RetGal maybe the method allowedToCanCopyFromThatPredecessorResource is no longer used and can be removed from the code?

RetGal:
@yvespp I can't find a method named "allowedToCanCopyFromThatPredecessorResource"

yvespp:
Ok, so the jsf pointed to method that doesn't even exist 😄

PR #196, opened 2017-09-04
Feature 9 jboss eap7 support

polandy:
Should now supports EAP7.
The Filter RequestParamViewHandler has been removed to avoid that all the RequestParams are always passed to the next page. This caused some Problems on EAP7.
Central Issues (which has been solved) when filter RequestParamViewHandler Filter has been removed:
- JSF Actions returned a String -> Destruction of the view scoped beans -> loss of the state after the execution actions
- in many cases the Parameter includeViewParams for the jsf links was set to "true" -> to many params has been gave over to the next page (a problem in editResourveView where the mask is used to edit a ResourceType and a Resource)
- Some params were not explicit passed to the next page
- in Settings some a4j did not work anymore. Changed those parts to normal jsf (h:commandLink,...)
- in Settings some Modal Panels did not work in the main form. They are located now in separate forms outside of the main form.


polandy:
Not implemented logview yet.

PR #199, opened 2017-09-24
Docker image

### LINKED ISSUE #82: provide docker image for liima
Liima is a deployment tool, but the deployment description of liima itself includes several manual steps (isn't it ironic?;)
- installing java 8
- installing old jboss 6.4 version
- installing jdbc driver
- installing war
- configuring security
- configuring data source

You should instead provide a docker image. The users of liima should not care about these low level tasks. It's way to complicated, if someone just want to give liima a try, without spending hours.

yvespp:
Push a docker image to dockerhub on every successful build via travis. If we want to run tests against the docker Image, I think it's easer to push the image instead of using the autobild from dockerhub. What do you think?
Repository is <https://hub.docker.com/r/liimaorg/liima/>
I created a separate account for the docker push, confirmation mail went to the puzzle-dev mailing list.
Closes #82


yvespp:
@RetGal or @phil-pona: what is the encryption key for the test database? I copied the key form the openshift config but I'm getting this error: `Caused by: java.security.InvalidKeyException: Invalid AES key length: 31 bytes`

yvespp:
Ok, found one that works in the test code here: https://github.com/liimaorg/liima/blob/fa7c1efd458882856a586ad920ddfeefdf5accc1/AMW_business/src/test/java/ch/puzzle/itc/mobiliar/business/generator/control/TemplateUtilsTest.java#L41

PR #200, opened 2017-09-25
Feature latest deployment job

### LINKED ISSUE #23: Deployment filter: Latest deployment job for ... shows two deployments for one env
If the deployment date of two deployments of the same app end env are exactly the same, the deployment filter "Latest deployment job for App Server and Env" show both deployments instead of one.

How to reproduce:
1. Create a deployment request for an AppServer with deployment date tomorrow at 12:00.
2. Create a second deployment request for the same AppServer with the same deployment date.
3. Select deployment filter "Latest deployment job for App Server and Env"

### LINKED ISSUE #42: Deploy filter: Latest deployment job for App & State filter
The combination of the filters "Latest deployment job for App Server and Env" and "State" doesn't show a deployment for the env if the state of the last deployment doesn't match the state filter.

In such a case, the most recent deployment that matches the filter should be shown.

The user wants to see only the latest deployment that was either failed or successful. States like canceled, rejected or requested should be ignored.

RetGal:
Fixes #42 and #23


yvespp:
@RetGal does #42 only affect the state filter or could other filters in combination with latest deployment job filter also be affected?

RetGal:
@yvespp It only affects the results if latest deployment job filter is set

yvespp:
Something is wrong here:
* Filter config: Application server: mos, Latest deployment job for App Server and Env shows 10 deployments (for each env one), on is failed the other 9 are success
* Add filter: State: success shows only one deployment but it should show 9

yvespp:
#23 doesn't seems to be fixed, the case in the issue still shows 2 deployments:
image

RetGal:
@yvespp the reported two issues are fixed now. There are two restrictions: Server side sorting does not work if Latest deployment job filter is set. If it is used in combination with the State filter, also the pagination does not work. Both wont be needed on the angular version of the page.

RetGal:
Fixes #206

yvespp:
@RetGal pull has merge conflicts

yvespp:
@RetGal while testing I noticed that now only one state filter can be set. Would it be hard to add this back?

RetGal:
@yvespp Multiple state filters can be added again.

Issue #75, opened 2017-04-04
Distinguishable deployment failure states


yvespp:
Currently it isn't obvious why a deployment failed. Users want to be able to see easily why a deployment failed to decide what action the should take. For example, if a deployment script failed it is very likely that server ist not running correctly and immediate action is required.

A deployment can fail on template generation (predeployment and deployment), predeployment script, deployment script, missing node, timeout, unexpected errors (runtime errors in Java), more?

Proposal:
* Add an additional field to the deployment entity: reason. This field has predefined values.
* Add a new deployment filter for the new field.
* Add a visual indicator on the deployment view why a deployment failed.
* Expose the new filed in the rest service.

RetGal:
@yvespp so far I got the following colors assigned to the different reasons:
- pre_deployment_script & deployment_script : red
- pre_deployment_generation & deployment_generation : orangered
- node_missing & timeout : darkorange
- unexpected_error & span.runtime_error : blue
Are these groups and colors appropriate?

RetGal:
TODO: Add this functionality to the angular deployment, as soon as this has been merged

yvespp:
I would keep red for all states and instead change the state text or even add an additional row to show the cause (if space allows). It should be still very obvious that a deployment failed.

I would not implement this for the old JSF GUI a all.

RetGal:
@yvespp ok, right now (jsf) looks like this
![failurereason](https://user-images.githubusercontent.com/19471064/30859480-3311ae26-a2c4-11e7-9047-3b90e0d40b38.png)

yvespp:
I would replace the status message (black text) with the text from the orange part and not show the orange part. On failure the box gets red as today but instead of failed shows the state it failed on. You know what I mean?

I think updating the JSF part is a bit of wast of time as we plan to remove it soon... You can always merge unfinished code if the feature is still disabled.

RetGal:
@yvespp like this?
![failure](https://user-images.githubusercontent.com/19471064/30863297-f3d265d6-a2d0-11e7-81b9-b135d2f80da4.png)

This feature requires a database change, while the bookmarkable_deployment_filter does not.

yvespp:
Like that but nicer ;-), the font is to small imho. Maybe we can make the state field bigger or make the text multi line?

RetGal:
@yvespp New screenshot in pr

PR #203, opened 2017-09-26
Feature deployment failure reason

RetGal:
Implements #75 (model and jsf part, todo:angular)


yvespp:
* The font of the state in the gui should definitely be bigger, I think we can increase the width of the state row.
* It would be nice if the reason doesn't contain the underscores.
* The reason should also be visible in the detail popup.

RetGal:
@yvespp
larger reason

Issue #214, opened 2017-11-02
Angular deployment screen review


yvespp:
1. The should screen show the last 10 deployments if no filters are set.
1. Results should be limited to 10 results per default, with the ability to increase it to 20, 50, 100.
1. Add paging. Without it, searches are very slow.
1. Use server side sorting, required for paging.
1. Deployments should be sorted by date by default (already the case).
1. Searching for "Application is *" doesn't not return any results.
1. Remove the wrapping box with title "Deployments" around the filters. Remove the box on the other angular screens to (create deployment, permission screen).
1. Auto apply filters when back button is hit or filter url is pasted into browser.
1. Filter "Latest deployment job for App Server and Env" together with "Environment" only shows one deployment for the env instead of one per App Server.
1. Add back Action icons form old deployment screen.
1. Add back progress animation form old deployment screen.
1. Remove modal animation: <https://v4-alpha.getbootstrap.com/components/modal/#remove-animation>
1. Clicking on an App in the deployment table gives a "A general System Failure occurred!": javax.el.ELException: /pages/editResourceView.xhtml @14,81 value="#{editResourceView.resourceIdFromParam}": java.lang.NullPointerException
1. Navigating to the deployment screen without any filters set in the url should should show the last filter settings.
1. Exception in log: `2017-11-02 14:46:16,942 logLev=ERROR class=org.apache.catalina.core.ContainerBase.[jboss.web].[default-host].[/AMW_rest].[ch.mobi.itc.mobiliar.rest.RESTApplication] thread="ajp-/0.0.0.0:8009-2" JBWEB000236: Servlet.service() for servlet
ch.mobi.itc.mobiliar.rest.RESTApplication threw exception: org.jboss.resteasy.spi.UnhandledException: javax.ejb.EJBTransactionRolledbackException: Unknown name value [pre_deployment_generation] for enum class [ch.puzzle.itc.mobiliar.busi
ness.deploy.entity.DeploymentFailureReason]`
1. ~~Rename "+ deployment" button to "Create deployment", "/ deployments" to "Edit deployments", capitalize the text of all buttons.~~ #215
1.  Hitting "Back to Deploy" from the log view shows no loader, same for 
1.  Canceling a deployment via "Edit deployment" shows `Error retrieving your data` (500 error via rest) instead of a user friendly error (deployment xy could not be canceled).
1. Add "failed" to the the failure reason e.g. "pre deployment generation failed"

yvespp:
[...]

PR #216, opened 2017-11-07
Feature bookmarkable filter enhancement

RetGal:
Includes all changes requested in #214, except 10.


yvespp:
Can you please change failure reason "timeout failed" to just "timeout"?

yvespp:
Maybe use "glyphicon glyphicon-repeat" instead of "glyphicon glyphicon-flash" for the redeploy action?

yvespp:
We could move stuff more to left now that the border is gone an there is now sidebar:
image

yvespp:
When I delete all Deployment filters the "Apply filter" button is deactivated. I think it should now be always active now that we show the last 10 deployment by default.

yvespp:
Clicking on the App server link takes me to the wrong AppServer. The id that is passed seams to be wrong.

RetGal:
@yvespp All done

yvespp:
Thanks!
Found an new one: pagination doesn't work with "Latest deployment job for App Server and Env" filter

PR #233, opened 2017-11-13
cleanup of AMW_db_scripts/pom.xml

yvespp:
Removed some unneeded stuff in the pom and updated doc.


yvespp:
Doesn't find the ora driver, move all db drivers to a different maven profile.

Issue #77, opened 2017-04-05
Config overview


yvespp:
The editResourceView currently show only the property values for the current environment. If you want to find out where a property is set you have to click through every env and check if it's green or not.
The missing overview leads users to accidentally overwriting values globally, because they think the are missing and don't realize that the values are defined on an different level or in the related resource.

Requirements:
* Get an overview of all properties that are defined on an app and relations of that app to check if everything is correct.
* Warn users if they are about to overwrite properties that are defined on a sub env.

PR #234, opened 2017-11-13
Feature 77 config overview

polandy:
- If a property is overriden in one of the sub-environments a warning icon appears next to the info icon of the property
- clicking on this warning icon opens a modal dialog on which all affected environments and values are listed
- it is implemented for properties of "normal", consumed- and provided resources
- if the resource is a related resource (consumed/provided) there is a warning that the properties of the child-environments will be overwritten


polandy:
closes #77

yvespp:
@polandy this pull has merge conflicts now. Sorry, we merge directly after it is resolved...

PR #247, opened 2017-11-15
removed passing the filter param to the log page. This is not necessa…

### LINKED ISSUE #243: AMW_angular/#/deployments: if filter is set, log view can not be shown and gives a http 500
The filter gets passed in the url to the log view: https://liima.mycorp.com/AMW_web/pages/logView.xhtml?deploymentId=317636&filters=[%20%20{%20%20%20%20%22name%22:%20%22Build%20success%22,%20%20%20%20%22comp%22:%20%22eq%22,%20%20%20%20%22val%22:%20%22true%22%20%20}]

If I remove the filter parameter (&filters*) the log view gets displayed correctly.

polandy:
* Could not reproduce the error. Anyway, it is not necessary to pass the filter to the log view and back because in the meantime the last filter is stored in the session.
* manually reverted 1fa293fe55f84b485e201498d6693932000b0c09
* fixes #243

### DIFF
--- a/AMW_angular/io/src/app/deployment/deployments-list.component.html
+++ b/AMW_angular/io/src/app/deployment/deployments-list.component.html
@@ -96,7 +96,7 @@
             </a><br>
           </ng-container>
           <ng-container *ngIf="deployment.actions.hasLogFiles">
-            <a class="action" href="/AMW_web/pages/logView.xhtml?deploymentId={{deployment.id}}&filters={{ filtersInUrl | json:0}}">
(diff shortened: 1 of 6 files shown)


yvespp:
I fear that with multiple browser tabs the filters get reset to the sessions stat, which might not be the filter that was set on the tab the user was coming from.
I analyzed the problem further and documented it in the issue. Can we maybe directly encode the url in the link to the log file?

yvespp:
Replaced by #252

PR #248, opened 2017-11-15
Clear filter button: clears the filter in the sessionStorage

### LINKED ISSUE #245: AMW_angular/#/deployments: filters come back
After a "Clear filter" and "Apply filter" go to a different page, if you come back to the deployment page, the filters are set again.
In this case, no filter should be set.

polandy:
fixes #245


polandy:
fixes  #245

yvespp:
btw the "fixes" and other keywords only work when they are in the first comment.

PR #266, opened 2017-11-21
Bugfix 246 avoid count for paging in rest call

### LINKED ISSUE #246: Deployment Filter latest Deployment slow
AMW_rest/resources/deployments?appServerName=test&offset=0&onlyLatest=true
Old time: 1-2 s 
New time: 12-14 s

polandy:
fixes #246, more infos [here](https://github.com/liimaorg/liima/issues/246#issuecomment-346085129)

### DIFF
--- a/AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
+++ b/AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
@@ -121,7 +121,7 @@ public class DeploymentsRest {
             sortingDirectionType = CommonFilterService.SortingDirectionType.valueOf(sortDirection);
         }
         LinkedList<CustomFilter> filters = createCustomFilters(filterDTOs);
-        boolean doPageingCalculation = maxResults > 0;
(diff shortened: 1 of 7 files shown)


AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
```diff
@@ -266,7 +267,7 @@ public Response getDeployments(
             createFiltersAndAddToList(DEPLOYMENT_PARAMETER, deploymentParameterValues, filters);
         }
 
-        Tuple<Set<DeploymentEntity>, Integer> result = deploymentBoundary.getFilteredDeployments(true, offset, maxResults, filters, null, null, null);
+        Tuple<Set<DeploymentEntity>, Integer> result = deploymentBoundary.getFilteredDeployments(false, offset, maxResults, filters, null, null, null);
```
yvespp: Can you do the same here as above instead of always disabling paging:
boolean doPageingCalculation = maxResults > 0

polandy:
With your requested change I've seen that it is not necessary to pass the parameter.

yvespp:
Ok, nice!

PR #270, opened 2017-11-22
Set cache max-age of one hour for non static content

### LINKED ISSUE #255: Angular and JSF caching Problem
I have had multiple reports of people that did not see the new Angular page and still so the jsf Page or they did see the new page but it was not displayed correctly. ctrl+alt+r in Chrome or ctrl+r in IE fixed this.

Also Changes that I made on the Angular deployment page in release 1.15.1 did not show up until I hit ctrl+alt+r in Chrome.

RetGal:
Fixes #255 for EAP7+ and WildFly8+. The bundles created by webpack must be reloaded after 1 hour (3600s)


yvespp:
@RetGal why only for EAP & Wildfly? Does it fix the caching of the base html as well?

RetGal:
@yvespp Yes, whit this approach, the whole angular part is cached for max 1 hour, then it has to requested again from the server (and reloaded if it does not get a 304). undertow-handlers are understood by Undertow only, that's why EAP7+ WildFly8+.

PR #280, opened 2017-12-05
Generate a list of all templates for resources

### LINKED ISSUE #231: Generate a list of all templates for resources
Currently it is only possible to access already generated templates via the ${templates} hash. Templates which are not generated at the time are not accessible. 
It should be possible to get a list of of all templates (Name, Path without Content) for all resources and all resources relations if possible grouped by resource type.

phil-pona:
fixes #231 
Ungenerated Templates on Resources can now be accessed via the ${template} hash

```
<#list templates?keys as key> 
${key} = ${templates[key].name}, ${templates[key].path}, ${templates[key].content}
</#list>
```

@Yvesp: do you have an idea how we want to access the relationtemplates


yvespp:
I would add them to the templates of the resource, e.g. `db.templates`.
Maybe add property that says that it is a relation template, e.g. `db.template.isRelationTemplate`.

PR #283, opened 2017-12-06
Reload properties after copy from resource

### LINKED ISSUE #222: editResourceView does not refresh correctly after "Copy from resource"
After "Copy from resource" the page is not refreshed and still shows the uncopied version of the resource.
After closing and reopening the resource it is displayed correctly.

RetGal:
Fixes #222


AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/copyResource/CopyResourceDataProvider.java
```diff
@@ -178,6 +182,7 @@ else if (copyFromResourceId == null) {
 			}
 			else {
 				GlobalMessageAppender.addSuccessMessage("Copy successful");
+				return true;
 			}
 		}
 		return false;
```
yvespp: Better remove `return true` at the end and move it into the if/else: compiler catches missing returns, else if doesn't have to be checked, the else if and else is not needed anymore:
 ```
		if (resourceToOverwrite == null) {
 			String message = "No resource selected.";
 			GlobalMessageAppender.addErrorMessage(message);
 			return false;
 		}
 	        if (copyFromResourceId == null) {
 			String message = "No resource to copy from selected.";
 			GlobalMessageAppender.addErrorMessage(message);
                        return false;
 		}
// remove else {
```

Issue #303, opened 2017-12-18
Delegation of permission feedback


yvespp:
Feedback to #74:
* The user should also get DEPLOYMENT ALL permission for that resource.
* The permission that get added should also be checked if the are granted already so not so many duplicates are created. Should use code from #289. Should also check permission from groups.
* Add a permission that decides if the permissions get added to the user if she/he creates a new resource. Name suggestion: ADD_ADMIN_PERMISSIONS_ON_CREATED_RESOURCE
* The permission GUI is not visible for a user with permission `PERMISSION_DELEGATION`.

RetGal:
@yvespp Yes, **every** newly created permission gets checked for possible duplicates. Which DEPLOYMENT permission (ACTION) shall he get?

DEPLOYMENT | CREATE | Darf Deployment Requests erstellen
-- | -- | --
  | READ | Darf die Deployment Seite sehen
  | UPDATE | Darf Deployment Requests confirment und das Datum updaten
  | DELETE | Darf Deployments rejecten

yvespp:
@RetGal DEPLOYMENT ALL limited to the resource.
The duplicate check doesn't check the permissions from roles then? I have all permissions in Liima from groups and still permissions get added to my user after I created an resource:
![image](https://user-images.githubusercontent.com/15231595/34103830-7b5459f2-e3ee-11e7-889c-e9336577b4b4.png)
How hard would it be to also check groups for duplicate/overlapping permissions?

I'm a bit worried that many permissions get added to users and it get's very hard to manage them (for example if you want to remove permissions from someone).

I think we should add a permission that decides if the permissions get added to the user if she/he creates a new resource. Name suggestion: `ADD_ADMIN_PERMISSIONS_ON_CREATED_RESOURCE`

yvespp:
@RetGal I'm uncertain about this features. I think we should not enable it for 1.16.
Can you please just add ADD_ADMIN_PERMISSIONS_ON_CREATED_RESOURCE permission so this feature doesn't change the behavior of Liima in 1.16?

I will think about how we want to go forward with it.

RetGal:
@yvespp In the meantime I already implemented the logic to check if a user already has a similar Restriction by granted by one of its Roles or directly by a UserRestriction. So I will add the ADD_ADMIN_PERMISSIONS_ON_CREATED_RESOURCE and only add the auto assigned permissions if the caller has it

RetGal:
[...]

PR #308, opened 2017-12-18
Feature permission delegation add deployment permission

RetGal:
Implements #303


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/security/control/PermissionService.java
```diff
@@ -850,6 +850,23 @@ public boolean hasPermissionToDelegatePermission(Permission permission, Resource
...
     }
 
+    /**
+     * Checks if the caller already has a Restriction similar to the given Restriction
+     * Returns true if he already has a similar, equal or less restrictive Restriction
+     * Returns false if he does not have a similar or just one that is more restrictive than the given one
+     *
+     * @param newRestriction
+     */
+    public boolean callerHasIdenticalOrMoreGeneralRestriction(RestrictionEntity newRestriction) {
```
yvespp: Why can't `public boolean identicalOrMoreGeneralRestrictionExists(RestrictionEntity newRestriction)` be used here?
The checks should be the same?
Same for `createAutoAssignedRestriction`.
Also one method is called `createSelfAssignedRestrictions` and the other `createAutoAssignedRestriction` -> stick to either self or auto Please.

RetGal: Not really, as the system does not know which Role/s a User has. `identicalOrMoreGeneralRestrictionExists` can only check if a similar Restriction of a User OR a Role exists. But the system knows which Role the logged in User (caller) has, so it can check if a User already has a similar Restriction, no matter if from one of its Roles or directly as UserRestricion.

Issue #295, opened 2017-12-13
Feedback #234 config overview


yvespp:
Here my feedback for #234:
* I don't get a warning when I'm overwriting a property on a related resource. A warning only appears if the another property is already overridden in a sub env. A warning should appear if a property in a relation is set that overrides values defined on the resource itself.

RetGal:
@yvespp Just to assure that I got you right: the warning on a property on a relation shall **only** appear, if it overrides a property defined on the (slave-) resource?

yvespp:
@RetGal 
For overwrite on resources it's ok like how it is.
For relation properties:
* For properties on global or domain env: show a read warning if it overwrites values defined on the resource. Ideally the a confirm dialog should appear when the user hits save which displays what get's overwritten.
* The config overview button should be shown if properties defined on the relation or on the resource are overwritten.
* The config overview should show what values are overwrite on the relation and on the resource.

PR #317, opened 2017-12-20
Bugfix config overview

RetGal:
Implements #295


yvespp:
@RetGal For relations:
* Can you also show if the property is defined on the relation or resource in the config overview?
* The warning `By changing the value of this property you will override the value of the following environments.` should only be showed, if properties defined directly on the resource are overwritten. Properties on a relation can't be overwritten on a parent env.
* Reword the warning: `Changing this property will override the values defined on the resource.`

PR #367, opened 2018-05-28
Trimming user and role names, added tests

### LINKED ISSUE #359: Angular permission gui: users and roles with spaces in name can be added
It is possible to add users and roles with trailing or leading spaces in the name.
Adding permissions to these users/roles doesn't add anything and the gui doesn't show any error or indication that it didn't work.
This is very confusing for users that accidentally added users/roles with trailing spaces.

RetGal:
Fixes #359


yvespp:
The rest interface should also refuse to crate roles with spaces.

PR #372, opened 2018-05-30
Enhancement create deployment with loader

### LINKED ISSUE #357: Angular create deploy: loader for new deployment
To avoid that a user accidentally clicks the create/request deployment button multiple times a loader should be shown after a click until the deployment was created.

RetGal:
Fixes #357.
Perhaps the loader is showed too often now.


yvespp:
The loader is no shown every time I change something on the GUI. Is it possible to show it only when the buttons "Request deployment" or "Deploy" are clicked?

Issue #343, opened 2018-03-26
REST error response body is not json encoded


dsbrng25b:
A lot of errors from the REST interface are not returned json encoded. e.g.:
* `curl -X GET "http://localhost:8080/AMW_rest/resources/deployments/not-existing-id/detail" -H  "accept: application/json"` (returns http 500 with a html encoded stack trace)
* `curl -X GET "http://localhost:8080/AMW_rest/resources/resources/validResource/validRelease/properties/notExistingProperty?env=Global" -H  "accept: application/json"` (returns html encoded not found error)

A good example is if you try to create a relation with a type which is not 'consumed' or 'provided':
```
curl -X POST "http://localhost:8080/AMW_rest/resources/resources/validResource1/validRelease/relations/validResource2?type=notexisting" -H  "accept: application/json"
```
this returns with http 400 and the following response body:
```json
{
  "message": "Type must either be 'consumed' or 'provided'",
  "detail": ""
}
```

Issue #347, opened 2018-04-09
REST get property endpoints do not return default values


dsbrng25b:
__Summary__
The REST endpoints to get properties do not return default values.
* /resources/{resourceGroupName}/{releaseName}/properties
* /resources/{resourceGroupName}/{releaseName}/properties/{propertyName}
* /resources/{resourceGroupName}/{releaseName}/relations/{relatedResourceGroupName}/{relatedReleaseName}/properties
* /resources/{resourceGroupName}/{releaseName}/relations/{relatedResourceGroupName}/{relatedReleaseName}/properties/{propertyName}

__Steps to reproduce:__ 
* create a resource type (e.g. myResource)
* add a property (e.g myProperty) with a default value (e.g. myDefault)
* create a resource (e.g myResource1) of the resource type created before
* get the property via REST /AMW_rest/resources/resources/myResource1/Past/properties/myProperty?env=Global

__Result:__
The request returns a value of 'null'.
```
{
  "name": "myProperty",
  "value": null,
  "replacedValue": null,
  "generalComment": null,
  "valueComment": null,
  "context": "Global"
}
```
__Expected Result:__
The request should return the default.
```
{
  "name": "myProperty",
  "value": "myDefault",
  "replacedValue": null,
  "generalComment": null,
  "valueComment": null,
  "context": "Global"
}
```

PR #403, opened 2018-06-26
Enhancement rest response body not json encoded

RetGal:
Fixes leftovers of #343


yvespp:
@RetGal do you mean #347?
`curl -H  "accept: application/json" -k http://localhost:8080/AMW_rest/resources/deployments/not-existing-id/detail -v` still returns html instead of json.

Issue #360, opened 2018-05-17
REST: add 'Test Generation' endpoint 


lbeyeler:
Add GET AMW_Rest/analyze/testGeneration/{resourceGroupName}/{releaseName}/{env}

Runs the test generation for the selected resource and returns

- 200 when the test generation is successful
- 500 when the test generation fails, the response body should contain the error logs

Thanks, LB

RetGal:
@lbeyeler Endpoint will be available unter `AMW_Rest/resources/analyze/testGeneration/{resourceGroupName}/{releaseName}/{env}` - as `resources` is the application path

PR #407, opened 2018-07-04
Feature test generation endpoint

RetGal:
Implements #360


yvespp:
Are the DTOs realy necessary?

habnux:
> Are the DTOs realy necessary?

Probably yes. It makes sense to return only the things which are required. Serializing entites might be sometimes be difficult too or even not possible without adding the right JSON annotations to them. In my opinion the added four DTOs are useful and correct in this case.

PR #416, opened 2018-07-16
REST: reload permissions & delete role

### LINKED ISSUE #414: ADD_ADMIN_PERMISSIONS_ON_CREATED_RESOURCE permission: only add DEPLOYMENT permissions if resource is an appServer
The following permissions should only be added on appServers:
- RESOURCE_TEST_GENERATION
- RESOURCE_TEST_GENERATION_RESULT
- DEPLOYMENT

### LINKED ISSUE #415: REST add permission: reload permission cache separately
A new REST methode should be added to reload the permission cache and the existing permission REST methode should get an option to not reload the cache. This can be used to delete and add new permissions without impacting users.

Also `createMultipleRestrictions` of the PermissionBoundary currently refreshes the [cache](https://github.com/liimaorg/liima/blob/66aac2595d02c219934397c056fa205a5d367de3/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/security/boundary/PermissionBoundary.java#L566) after adding a single permission which is not necessary and slows it down.

yvespp:
Closes #415 
Closes #414 

PermissionBoundary
* public methods got a new reload parameter
* private methodes no longer reload the cache
* createAutoAssignedRestrictions only adds valid permissions for the resourceType
* new method delete role, cascade deletes restrictions as well

RestrictionsRest
* nutating methods got an optional reload parameter, default true
* new reload method
* new method delete role

Other:
* added new catch all exception mapper
* fixed some mapper


yvespp:
@RetGal I'm goning to merge this without you review. Please look at it anyway, will fix it in a follow up.

PR #444, opened 2018-12-03
Fixed #442

### LINKED ISSUE #442: Multiple releases of Runtime displayed wrong on Appserver
Multiple releases of a Runtime are displayed as multiple relations on the Appserver:
image

In the screenshot the runtime EAP70 has two releases. Only one relation should be displayed.

RetGal:
Fixes #442


RetGal:
@yvespp done (including update of surefire and jacoco versions)

Issue #441, opened 2018-10-15
Custom properties on environement


yvespp:
We want to add alternative/new environment names to an existing env. These additional info should be available like env name and domain name in templates (`${env.name}`, `${env.domain}`}. Example: `${env.myCustomProp}`

It seems the Liima data model already has the needed DB tables for it but that has to be verified. It also has to be check if the generator adds these properties to the template model correctly.
The GUI for the properties could be added to the environment setting page (`AMW_web/pages/settings.xhtml?mode=envs`).

yvespp:
If a generic solution is to complicated, a fixed field could be added to the env entity directly.

PR #447, opened 2018-12-09
Added feature and test

RetGal:
Implements #441 => ${env.nameAlias}
I had to downgrade H2 from 1.4.97 to 1.4.96 due issues with Liquibase. 
See https://github.com/h2database/h2database/issues/1073


yvespp:
Can you make the nameAlias show up in the GUI next to the environment (in brackets)? Forgott to specify this in the issue...
I think the folowing pages are affected: editResourceView, Angular Deployment and create Deployment screen. Maybe the rest interface has to be updated to include the new field as well?
Thanks for your work!

PR #456, opened 2018-12-21
Fixes REST: put properties call very slow

### LINKED ISSUE #446: REST: put properties call very slow
The REST Call to set a value on a property of a resource is very slow. Could you please check and optimize this ?

Call :
/resources/resources/<resource_name>/<release_name>/properties/<property_name>

Thanks in adance
Marc Wenger (mobi)

yvespp:
Fixes #446
Remove REQUIRES_NEW transaction attributes on context and resource repositories so entities aren't detached and don't need to merged again.
The merges caused about 5k SQL queries. The request went from 20s to under 1s with this change.

Why did we put REQUIRES_NEW on most boundries? Imho entities should not be detached when they are passed around inside the business logic.


RetGal:
@yvespp REQUIRES_NEW expresses that the transaction is controlled by the boundary which starts a new transaction explicitly. I.m.h.o. this makes sense on boundaries which are *exclusively* used from UI components.

PR #462, opened 2019-03-13
Enhancement performance edit reseource view

RetGal:
Performance improvements for the edit resource view. 
- added optimized query for Oracle
- removed SoftlinkRelation from WEB


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/property/control/PropertyEditingQueries.java
```diff
@@ -282,4 +274,16 @@ private String loadSQLFile(String file) {
             throw new RuntimeException(e);
         }
     }
+
+    private boolean isH2() {
```
yvespp: Duplicate Code
https://github.com/liimaorg/liima/blob/0daea903e6b6a0962b21870957f3f44aabafe9bf/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/domain/applist/ApplistScreenDomainServiceQueries.java#L90

PR #463, opened 2019-03-14
Enhancement performance indexes

RetGal:
Added changesets for creating additional indexes. Does not add indexes for H2 and Oracle XE.


yvespp:
@RetGal Why are the indexes not created for H2 and Oracle XE?

RetGal:
> @RetGal Why are the indexes not created for H2 and Oracle XE?

Both do not support the online option.

PR #470, opened 2019-03-19
Fix indexes syntax

yvespp:
Update Liquibase version and schema, remove online option, run on all db systems


yvespp:
Without the `online` option no DML on the table is possible during index creation, which is not needed 
here.

PR #485, opened 2019-05-08
deployment queries: only set optimizer if pageing is used

yvespp:


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentBoundary.java
```diff
@@ -196,6 +196,7 @@ public boolean isPossible() {
 
         DeploymentState lastDeploymentState = null;
         boolean hasLastDeploymentForAsEnvFilterSet = isLastDeploymentForAsEnvFilterSet(filters);
+        boolean setOtimizer = doPaging && !hasLastDeploymentForAsEnvFilterSet && dbUtil.isOracle();
```
RetGal: set**Optimizer**

Issue #484, opened 2019-05-08
IllegalArgumentException after new release of relation


yvespp:
If a new release of a related resource is created where there is an instance property set from the consumer an IllegalArgumentException is throw if that new relation is displayed in the editResourceView.

How to reproduce:
1. Create new resource resA of type application
1. Create new resource resB of any type
1. In resB add an instance property
1. In resA add resB as a consumed relation
1. In relation form resA to resB set a value in the new instance property
1. Create a new release of resB

Result: if you try to open the relation from resA to resB of the new release the exception below is thrown.

JSF:
```
editResourceView.xhtml "#{editResourceView.resourceIdFromParam}": java.lang.IllegalArgumentException
```

Exception from server log:
```
2019-05-08 14:26:14,030 logLev=ERROR class=org.jboss.as.ejb3.invocation thread="default task-163" corrID= caller= user= WFLYEJB0034: EJB Invocation failed on component PropertyEditor for method public java.util.List ch.puzzle.itc.mobiliar.business.property.boundary.PropertyEditor.getPropertiesForRelatedResource(java.lang.Integer,ch.puzzle.itc.mobiliar.business.property.entity.ResourceEditRelation,java.lang.Integer): javax.ejb.EJBException: java.lang.RuntimeException: java.lang.IllegalArgumentException
        at org.jboss.as.ejb3.tx.CMTTxInterceptor.invokeInOurTx(CMTTxInterceptor.java:246)
        at org.jboss.as.ejb3.tx.CMTTxInterceptor.required(CMTTxInterceptor.java:362)
        at org.jboss.as.ejb3.tx.CMTTxInterceptor.processInvocation(CMTTxInterceptor.java:144)
        at org.jboss.invocation.InterceptorContext.proceed(InterceptorContext.java:422)
        at org.jboss.invocation.InterceptorContext$Invocation.proceed(InterceptorContext.java:509)
        at org.jboss.weld.module.ejb.AbstractEJBRequestScopeActivationInterceptor.aroundInvoke(AbstractEJBRequestScopeActivationInterceptor.java:72)
        at org.jboss.as.weld.ejb.EjbRequestScopeActivationInterceptor.processInvocation(EjbRequestScopeActivationInterceptor.java:89)
        at org.jboss.invocation.InterceptorContext.proceed(InterceptorContext.java:422)
[...]

PR #486, opened 2019-05-10
Switch back to unoptimized query for oracle

yvespp:
Switch back to unoptimized query for oracle and add logging for property creation exceptions.

With this change related instance properties don't cause an exception anymore. See #484.
The underlying issue is not fixed with this, but it restores the app to a working state.


yvespp:
Reverted query introduced with https://github.com/liimaorg/liima/pull/462

PR #503, opened 2019-08-20
removing slf4j since it is duplicated

sdedijer:


yvespp:
Duplicated where? I found it only once in the source code...

sdedijer:
Hi Yves, 

Once I run AMW_business tests, then I got following warning:
`SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/sdedijer/.m2/repository/ch/qos/logback/logback-classic/1.2.3/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/sdedijer/.m2/repository/org/slf4j/slf4j-log4j12/1.7.7/slf4j-log4j12-1.7.7.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]`

Then when I checked dependency:tree for slf4j (`mvn dependency:tree -Dverbose -Dincludes=org.slf4j`) in the project I saw that there are transitive dependencies already, so I removed this direct one. After that tests were running without warning. Could you also please try it all on your side?
And of course I can revert my changes, no problem at all, just please let me know. It was in any case just warning that appeared, tests were running further on in both cases.

Thanks and BR,
Sanja

Issue #418, opened 2018-07-23
Environment Strings in the REST API are case sensitive


wemu:
Hello

I notived when using the Rest API that passing the environment string into the API the case must match exactly the one stored in liima. For example using the Deployments Filters API:

`[{"name":"Environment","comp":"eq","val":"t"},{"name":"Application server","comp":"eq","val":"jap_mobitor"}]`

does return nothing but:

`[{"name":"Environment","comp":"eq","val":"T"},{"name":"Application server","comp":"eq","val":"jap_mobitor"}]`

does. If environment String contain more than one letter this gets very picky.

I'd like to have the environment paramters to be case-insensitve on the Rest API

PR #504, opened 2019-08-21
Enhancement envsearch case insensitive

lackrobin:
Fixes issue #418
Making the filter for Environments case-insensitive


AMW_business/src/test/java/ch/puzzle/itc/mobiliar/business/domain/commons/CommonFilterServicePersistenceTest.java
```diff
@@ -139,4 +146,62 @@ public void test_addSpecialAndJoiningFilterAndCreateQuery(){
         assertThat(query.getParameters().iterator().next().getName(), is("Deploymentparameter0"));
     }
 
+
+    @Test
+    public void test_environmentCaseInsensitiveSearchUpper(){
```
yvespp: Only two lines of code differ between the two tests. Maybe move the code to a separate function and call it with different parameters. It's easier to read and maintain.

yvespp:
Thanks!

PR #505, opened 2019-08-21
Bugfix multiple hostnames

### LINKED ISSUE #472: Host in multiple Releases even if not the case
Since Runtimes have several Releases, when you search a Hostname that belongs only to 1 AppSrv/Release, Liima indicates "multiple" in the "AppServer Release" Column.

Moreover, the REST Call "hostNames" seems to have the same bug : 
If you execute the call by specifying a"host" (which is only configured once in Liima) and "disableMerge=true", liima send back 2 times the same data (ServerTuple) in the answer.

RetGal:
Fixes #472


yvespp:
@RetGal can we filter this via unique query or modify the query so it only fetches one runtime and not all?

RetGal:
@yvespp not really, besides the Runtime mentioned in the issue, also the Node Release was buggy

PR #574, opened 2020-10-06
Migration to Boostrap 4, replace datepicker

### LINKED ISSUE #510: Replace eonasdan-bootstrap-datetimepicker
See #483
The datetimepicker is no longer maintained, see: https://github.com/Eonasdan/bootstrap-datetimepicker
Requires a separate version of jquery.

### LINKED ISSUE #570: Date icon doesn't open calendar
Clicking on the date icon of the deployment filter "Deployment date" doesn't open the calendar anymore:
deploy_filter
Clicking inside the field opens it however.
If there are two deployment filters with a date field, the calendar of the second filter can't be opened at all.

More:
- Changing the date of a existing deployment by clicking on the deployment date only opens the calendar if you click inside the field.
- If you edit deployment via "Edit" button and select "Change date" it works correctly.
- In the "Create new deployment" screen it still work by clicking on the calendar icon of the "Deploy at" field.

Maybe we should fix it by implementing #510

mburri:
This MR contains:
- migration from bootstrap 3 to bootstrap 4
- scss cleanup
- new icons (bootsrap icons instead of glyphicons - the glyphicons set is no longer available for bootstrap)
- replacement of the existing eonasdan-boostrap-datetimepicker with a custom datetimepicker component based on ng-boostrap (closes #510)
- closes #570 

The bootstrap migration was necessary because of ng-boostrap/ the ng-bootstrap datepicker and timepicker component that depend heavily on bootstrap 4 - and did not properly work when part of the application was still based on bootstrap 3

a lot of custom scss/css could be deleted as a benficial side effect which improves the frontends overall maintainability


AMW_angular/io/package.json
```diff
@@ -24,10 +27,10 @@
     "@angular/router": "~10.0.8",
     "@ng-bootstrap/ng-bootstrap": "^7.0.0",
     "@ng-select/ng-select": "~5.0.1",
-    "bootstrap-3-typeahead": "4.0.2",
-    "bootstrap-sass": "3.4.1",
-    "eonasdan-bootstrap-datetimepicker": "4.17.47",
+    "bootstrap": "^4.5.0",
+    "bootstrap-icons": "^1.0.0",
     "jquery": "^3.5.1",
```
yvespp, line 32: Should we be able to remove jquery now?

mburri: no - there are still some modal dialogs that directly use jquery - these could be changed (and I'd be happy to do it) but I think that should go into another issue / pr

mburri:
Thanks for your feedback - I fixed the issues you mentioned - but I'm not sure about the styling of the buttons - do you prefer the ones with the outline (deployments page) or the the ones with the light background (e.g. the cancel button on permissions -> role -> edit)

yvespp:
Personally I like the buttons without boarder/light background better but the contrast could be better. On my screen I can barely see where the button starts. Also active and passive buttons are hard to keep apart: 
image

I noticed some more stuff:
* On create deployment can you aligne the buttons with the text:
image
* The (i) in the deployment state of the deployments table doesn't show the clickable/hand icon on mouse over.
image
* Can you make the navbar fixed again so it stays at the top of the page all the time?
* The font size is lager now but I think that's ok...

yvespp (approved):
Thanks!

PR #586, opened 2021-01-06
Feature/450 improve deployment log

### LINKED ISSUE #443: Angular GUI for deployment log view
The log view of the deployment page is still JSF and should be replaced with Angular.
Currently the Angular SPA has to be loaded again and again if you switch between log and deploy page.

### LINKED ISSUE #450: Make errors more visible in deployment log
To make the Liima Log more readable errors and causes for deployment failure should be highlighted.

Ideas:
* Color logs to std error differently
* Highlight lines that container error/err ors some specific code. Could maybe done via CodeMirror Langugage mode: https://codemirror.net/doc/manual.html#modeapi
* Use a different JavaScript lib to display the log

mburri:
- [x] new rest endpoint to fetch available log files from backend
- [x] implement deployment log view with angular
- [x] delete legacy view (jsf) and bean
- [x] highlight some keywords with codemirror - more keywords possible
- [x] make file selection a little bit more obvious

fixes #450 and also fixes #443


mburri:
The lates commit highlights whole lines where certain keywords appear:
image

yvespp:
@mburri thanks for the work so far. Here my feedback:
* Can you please add some space between the log file selector and the log box?
* In my chrome the logfile box doesn't use the whole screen vertically. Can you make it so it uses the whole screen height? image
  * Maybe the gray box around the log could be removed and the scrollbar of the browser could be used instead?
* Maybe the title `Deployments` could be removed and replace with `Log file for 709104`
* Can you turn of the line numbering and enable the search like in the old log view?

mburri:
- [x] added some space between file selector and codemirror box
- [x] use full screen height
- [x] changed page title
- [x] turn of line numbering
- [x] enable codemirror search box
- [x] fixed markup to remove grey box around log and use the browsers sroll bar
  
image

yvespp:
Can make it look like this please?
image

yvespp:
Somehow the log box still doesn't use the whole height of the screen:
image

Tried in chrome and IE.

AMW_angular/io/src/app/deployments/deployments-list.component.html
```diff
@@ -75,7 +75,10 @@
       </tr>
     </thead>
     <tbody>
-      <tr *ngFor="let deployment of deployments" class="deploy-{{ deployment.state }}">
+      <tr
```
yvespp, line 78: For me this formatting is harder to read... does you IDE automatically do this?

yvespp (changes requested):
For the rest service:
I think it would be better if `/deployments/{id}/logs` would only return the info which logs exist and what their names are (only metadata). The logs can then be received with `/deployments/{id}/logs/{logid/logName}`.
Why:
* The log file doesn't have to be escaped to fit into json.
* `/deployments/{id}/logs/{logid}` can return just plain text and a client doesn't have to be able to read json to read the log.
* The log view must only load the file that is visible.

AMW_angular/io/src/app/deployments/deployments-list.component.html
mburri: I'm using prettier as code formatter, yes. 
what's the problem exactly:
- that attributes are placed on a new line?
- the closing bracket that is on a new line?

yvespp: That the line is split into multiple lines even though it would fit on one line.

mburri:
as requested: 

- [x] raise max. number of characters per line (prettier)
- [x] changed backend api (get log file content as text/plain
- [x] highlight warning messages
also improved:
- [x] set log file name the url to allow direct links

yvespp (approved):

yvespp:
Thanks for you work and patience!

PR #587, opened 2021-01-14
Convert modals to Ngb

yvespp:
* Converts all modals to Ngb
* Removes all jquery code. @mburri can we remove it from dependencies now?
* Datetime deployment filter is stored as string again in the url
* Fix time format usage


mburri:
I fixed the failing test and fixed a small code smell.
Trying to also remove the jquery dependency now...

mburri:
unfortunately - boostrap itself uses jquery as a dependency - but it will go away with bootstrap 5 that is currently in beta

mburri (approved):

AMW_angular/io/src/app/deployment/deployment.component.spec.ts
```diff
@@ -327,7 +285,7 @@ describe('DeploymentComponent (create deployment)', () => {
       id: 5,
       tagDate: 1485378084103,
     } as ResourceTag;
-    component.deploymentDate = DateTimeModel.fromLocalString('02.01.2017 12:00');
+    component.deploymentDate = DateTimeModel.fromLocalString('02.01.2017 12:00', DATE_FORMAT_ANGULAR);
```
yvespp, line 288: I think DATE_FORMAT should be used here, as the model uses moment to parse the date and DATE_FORMAT_ANGULAR is not valid for it: https://github.com/liimaorg/liima/pull/587/commits/4802703a7bb3aaf1c6cb08c92773d4742e3f6717#diff-f3840bdde8f725c868fdc60e05932fd5773a8648c00d0af6e50dab596d42de5aR3

PR #590, opened 2021-01-21
Issue/487 copied instance properties

### LINKED ISSUE #487: New release of relation uses old prop descriptor for copied instance property
Most likely this is the underlying cause for #484.

How to reproduce:
1. Create new resource resA of type application
1. Create new resource resB of any type
1. In resB add an instance property. Take note of the descriptor id.
1. In resA add resB as a consumed relation
1. In relation form resA to resB set the value to `myUniqueValue123` in the new instance property
1. In the DB search for the new value: `select * from tamw_property where to_char(substr(value, 0, 30)) = 'myUniqueValue123';`
  The descriptor_id should be the same as above.
1. Create a new release of resB

=> search the DB again. It should have two values with two different deployment descriptor ids as the copy created a new resource with a new descriptor.
But what happens is that the two values have the same descriptor id, meaning the value is attached to the property descriptor of the old release instead of the new one.
This then causes the optimized oracle query to fetch wrong data which in turn causes an exception. The old query is immune to this.
If you set the a new property value in the relation from resA to resB it then gets attached to correct descriptor.

If I change the descriptor id of the new property to the descriptor id of the new release in the DB, then the optimized query works and also shows the copied value.

This bug seems to have been in Liima for a long time....

mburri:
Commit 48180f1 fixes the issue: a new property descriptor is created and persisted if the property belongs to the copied resource.

I started to refactor the method afterwards to clean up the code an make the intent more clear. Most refactorings were performed using the extract method/ invert "if" confition refactorings from intellij.

The existing (parametrized) tests already covered almost all affected code - I had to add only one for the case where the value of the origin property is merged into the target property.

fixes #487


yvespp:
I tested this on our env and some values are now not copied when:
* Descriptor is defined on type
* The property value is defined on the copied resource.

The same happens for properties that are defined on the relations of the copied resource.

yvespp:
Replaced by #659

PR #592, opened 2021-01-22
434 template paths

### LINKED ISSUE #434: Prevent that templates can be writen outside of generation directory
We should check that Liima templates can only be writen to the generation directory. 
Especially to prevent path traversel attacks: <https://en.wikipedia.org/wiki/Directory_traversal_attack>
It could be that this is already prevented.

We could prevent a leading `/` and `..` in template paths all together or resolve a relative template path to an absolute path and then check that it is inside the generation dir.

mburri:
Validate the specified path for the template. Absolute paths starting with "/" and path traversals "../" are not allowed.

I allowed myself to add some wrapper methods in order to be able to write some unit tests and slightly refactor the save() method (mostly for readability)

fixes #434

### DIFF
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/template/boundary/TemplateEditor.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/template/boundary/TemplateEditor.java
@@ -195,10 +195,24 @@ public class TemplateEditor {
 		return permissionService.hasPermissionToUpdateResourceTemplate(res, testingMode);
 	}
 
-    void saveTemplate(TemplateDescriptorEntity template, HasContexts<?> hasContext) throws AMWException {
-		if (StringUtils.isEmpty(template.getName())) {
+	void validateTemplate(TemplateDescriptorEntity templateDescriptorEntity) throws AMWException {
+		if (StringUtils.isEmpty(templateDescriptorEntity.getName())) {
 			throw new AMWException("The template name must not be empty");
 		}
+
+		if (templateDescriptorEntity.getTargetPath() != null && templateDescriptorEntity.getTargetPath()
+																						.startsWith("/")) {
+			throw new AMWException("Absolute paths are not allowed for file path");
+		}
+
+		if (templateDescriptorEntity.getTargetPath() != null && templateDescriptorEntity.getTargetPath()
+																						.contains("../")) {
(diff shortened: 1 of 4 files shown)


yvespp:
@mburri would it be better to move this validation to the business module? Else it has to be reimplemented for rest/angular again.

mburri:
@yvespp moving the logic to business-module makes sense.
I'm not sure about the best way to handle the permission check. There is one in the EditTemplateView itself: https://github.com/liimaorg/liima/blob/5429171e7e802f0541e4e83ac4de52f1a3b90356/AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/templateEdit/EditTemplateView.java#L266

And one in the bussiness-module but only for templates for relations:
https://github.com/liimaorg/liima/blob/master/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/template/boundary/TemplateEditor.java#L134

There is no permission check in the TemplateEditor for the other possible types (ResourceType/ Resource), see: https://github.com/liimaorg/liima/blob/5429171e7e802f0541e4e83ac4de52f1a3b90356/AMW_web/src/main/java/ch/puzzle/itc/mobiliar/presentation/templateEdit/EditTemplateView.java#L276

I tried to move these to the template editor (different branch) but abandoned it for now because it became a really big change...

I think the simplest change would be to keep the permission check in EditTemplateView and move the template path validation to the TemplateEditor - or what do you think?

mburri:
I moved the permission checks for resources and resource types to the template editor (see branch 434-template-paths-permission-check-explorations) - but i got quite confused by the method TemplateEditor#saveTemplateForRelation that already does check permissions - but in a different way than EditTemplateView#canModifyTemplates()

Further investigations revealed, that the permission checks to create/ update templates for resource, resourcetypes and/ or resource relation (that seem to fall back to resources or resource types) are also covered by annotations. The specific implementations are necessary to distinguish between create/ update a template.

What is the preferred way for permission checks in Liima:
- Annotations in the business-module
- Custom Logic/ throwing Exceptions in the business-module
- Custom Logic/ throwing Exceptions in the web-module (see EditTemplateView#canModifyTemplates())

Personally, I'd like to use annotations where possible.

yvespp:
Via annotations is fine. For complicated checks it's probably easier in code.

mburri:
I'm closing this issue without merging in favor of #595

PR #595, opened 2021-03-01
434 template paths permission check explorations

mburri:
With this pull request, the permission checks to save/ update a template for resources, resource types and resource (type) relations is completely pushed to the business module.

Template Paths are now validated and absolute paths and paths traversals are not allowed

### DIFF
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/security/control/PermissionService.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/security/control/PermissionService.java
@@ -30,8 +30,8 @@ import ch.puzzle.itc.mobiliar.common.exception.NotAuthorizedException;
 import ch.puzzle.itc.mobiliar.common.util.DefaultResourceTypeDefinition;
 import org.apache.commons.lang3.StringUtils;
 
-import javax.ejb.SessionContext;
 import javax.ejb.Schedule;
+import javax.ejb.SessionContext;
 import javax.ejb.Stateless;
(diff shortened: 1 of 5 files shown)


yvespp:
I tested this and works as expected, so I think we should go forward whit this.
Sometimes the code format is a bit odd, like here: https://github.com/liimaorg/liima/pull/595/files#diff-feed02a064a2d604d60b02b8b980888d34d72994ebf6353efd733c0b42ae8e77R152
But could only be a GitHub issue.

Maybe you could re add the removed tests on the business side?
Can you close #592 then?

mburri:
@yvespp 
I implemented the tests - and noticed that the permission checks where somehow flawed:
if the testing mode flag was set to true, and the user has Permission.SHAKEDOWN_TEST_MODE - it was still checking for Permission.RESOURCE_TEMPLATE with the proper action.

I changed that: for testing mode, only Permission.SHAKEDOWN_TEST_MODE is requested - is this correct? Or does the user still need Permission.RESOURCE_TEMPLATE or Permission.RESOURCETYPE_TEMPLATE when in testing mode?

yvespp:
@mburri I think only checking for SHAKEDOWN_TEST_MODE is ok.
SHAKEDOWN_TEST_MODE is a global permission so with the change a user can change shakedown templates on all resources instead of only the ones where he/she also can change templates. But I think that's fine because SHAKEDOWN_TEST_MODE is more of an admin permission and the shakedown functionality is hardly ever used an probably should be removed...

yvespp (approved):

PR #598, opened 2021-03-16
replace momentjs with date-fns

srothPuzzle:
Replaced momentjs with date-fns

### DIFF
--- a/AMW_angular/io/package.json
+++ b/AMW_angular/io/package.json
@@ -29,13 +29,13 @@
     "@ng-bootstrap/ng-bootstrap": "^8.0.4",
     "@ng-select/ng-select": "^5.1.0",
     "bootstrap": "^4.6.0",
-    "bootstrap-icons": "^1.3.0",
-    "codemirror": "^5.59.2",
-    "jquery": "^3.5.1",
-    "moment": "^2.29.1",
-    "rxjs": "~6.6.3",
-    "tslib": "^2.0.0",
-    "zone.js": "~0.11.3"
+    "bootstrap-icons": "^1.4.0",
+    "codemirror": "^5.59.4",
+    "date-fns": "^2.19.0",
+    "jquery": "^3.6.0",
+    "rxjs": "^6.6.6",
+    "tslib": "^2.1.0",
+    "zone.js": "~0.10.3"
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "~0.1101.2",
@@ -59,6 +59,6 @@
     "protractor": "~7.0.0",
     "ts-node": "~8.10.2",
     "tslint": "~6.1.3",
-    "typescript": "~4.0.5"
+    "typescript": "^4.0.7"
   }
 }
--- a/AMW_angular/io/package.json
+++ b/AMW_angular/io/package.json
@@ -35,6 +35,7 @@
     "jquery": "^3.6.0",
     "rxjs": "^6.6.6",
     "tslib": "^2.1.0",
+    "typescript": "~4.1",
     "zone.js": "~0.10.3"
   },
   "devDependencies": {
@@ -59,6 +60,6 @@
     "protractor": "~7.0.0",
     "ts-node": "~8.10.2",
     "tslint": "~6.1.3",
-    "typescript": "^4.1.5"
+    "typescript": "~4.1"
   }
 }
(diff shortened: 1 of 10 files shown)


AMW_angular/io/src/app/core/amw-constants.ts
```diff
@@ -1,5 +1,5 @@
 export const AMW_LOGOUT_URL = 'amw.logoutUrl';
 // used for moment.js
-export const DATE_FORMAT = 'DD.MM.yyyy HH:mm';
+export const DATE_FORMAT = 'yyyy-MM-dd_HHmm';
 // use by angular date pipes (unicode): https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table
 export const DATE_FORMAT_ANGULAR = 'dd.MM.yyyy HH:mm';
```
yvespp, line 5: I think DATE_FORMAT_ANGULAR is no longer required and can be removed

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -364,7 +364,7 @@ export class DeploymentsComponent implements OnInit {
   private pushDownload(prefix: string) {
     this.isLoading = false;
     const docName: string =
-      prefix + '_' + moment().format('YYYY-MM-DD_HHmm').toString() + '.csv';
+      prefix + '_' + datefns.format(new Date(), 'yyyy-MM-dd_HHmm').toString() + '.csv';
```
yvespp, line 367: Here you could use DATE_FORMAT from amw-constants

AMW_angular/io/package.json
```diff
@@ -59,6 +59,6 @@
     "protractor": "~7.0.0",
     "ts-node": "~8.10.2",
     "tslint": "~6.1.3",
-    "typescript": "~4.0.5"
+    "typescript": "^4.0.7"
```
yvespp, line 62: I think this should be change to `~4.1`
Should fix https://github.com/liimaorg/liima/runs/2119744614?check_suite_focus=true#step:7:555

Angular supported version: https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3
node-semver: https://github.com/npm/node-semver#caret-ranges-123-025-004

AMW_angular/io/src/app/core/amw-constants.ts
srothPuzzle: It is used in auditview-table.service.ts but only once.

yvespp: Can you please refactor it so that only `DATE_FORMAT` is used and delete `DATE_FORMAT_ANGULAR`?

AMW_angular/io/src/app/shared/date-time-picker/date-time.model.ts
```diff
@@ -18,60 +18,61 @@ export class DateTimeModel implements NgbDateTimeStruct {
     Object.assign(this, init);
   }
 
-  private static fromMoment(m: moment.Moment): DateTimeModel {
-    if (!m.isValid()) {
+  private static fromMoment(date: Date): DateTimeModel {
```
yvespp, line 21: This method needs to be renamed to `fromDate`

AMW_angular/io/package.json
```diff
@@ -29,13 +29,14 @@
...
-    "rxjs": "~6.6.3",
-    "tslib": "^2.0.0",
-    "zone.js": "~0.11.3"
+    "bootstrap-icons": "^1.4.0",
+    "codemirror": "^5.59.4",
+    "date-fns": "^2.19.0",
+    "jquery": "^3.6.0",
+    "rxjs": "^6.6.6",
+    "tslib": "^2.1.0",
+    "typescript": "~4.1",
```
yvespp, line 38: I don't think typescript is needed here

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -364,7 +365,7 @@ export class DeploymentsComponent implements OnInit {
   private pushDownload(prefix: string) {
     this.isLoading = false;
     const docName: string =
-      prefix + '_' + moment().format('YYYY-MM-DD_HHmm').toString() + '.csv';
+      prefix + '_' + datefns.format(new Date(), DATE_FORMAT).toString() + '.csv';
```
yvespp, line 368: @srothPuzzle this is not the same format as `DATE_FORMAT`
Can you please change it to `yyyy-MM-dd_HHmm`.

yvespp (approved):
Thanks!

PR #651, opened 2022-04-02
Only delete specific resource release

### LINKED ISSUE #623: REST DELETE: /resources/resourceGroups/{resourceGroupId}/releases/{releaseId} deletes non matching releases
The REST DELETE will delete the next available releases of resources if the resource release doesn't exist for the releaseId.

1. Create a resource with two releases: release1 and release2
2. release2 is after release1
2. Delete release2 via REST-> the correct release will be removed
3. Delete release2 via REST again -> release1 will be removed and the resource group will be removed as well.

RetGal:
Fixes #623 @yvespp this is a private open source contribution :)


yvespp:
Hi @RetGal, thanks for you contribution! 👍 

Couldn't we use this method instead of extending `getResourceEntityForRelease`:
https://github.com/liimaorg/liima/blob/master/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/resourcegroup/boundary/ResourceLocator.java#L138
We already have a lot of methods that are very similar and I would like to avoid adding another one...

RetGal:
@yvespp yes, you're right: see https://github.com/liimaorg/liima/pull/652

PR #663, opened 2022-10-13
Feature/validate column to sort

dowhiletrue:


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentBoundary.java
```diff
@@ -1551,4 +1554,16 @@ public FileVisitResult postVisitDirectory(Path dir, IOException e) throws IOExce
             }
         }
     }
+    private Sort createSort(String colToSort, Sort.SortingDirectionType sortingDirection, boolean lowerSortCol) {
+        Sort.SortBuilder sortBuilder = Sort.builder();
+        DeploymentFilterTypes.DEPLOYMENT_FILTER_TYPES_FOR_ORDER.stream()
+                .map(DeploymentFilterTypes::getFilterTabColumnName)
+                .filter(c -> Objects.equals(c, colToSort))
```
yvespp, line 1561: Wouldn't it be better to move the validation into the Sort class? Some code could by accident still pass a bad/unvalidated string to it which then ends up in the query.

yvespp (approved):

Issue #653, opened 2022-04-27
Add audit tables for permissions entities


yvespp:
Currently the permission entities are not audited. Because the security relevant the should be audited. Affected entities: https://github.com/liimaorg/liima/tree/master/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/security/entity

yvespp:
Closed by #653

PR #665, opened 2022-10-18
Issue/622 avoid duplicate releases

dowhiletrue:
* constraint currently results in server error (500). If this is not adequately good for the circumstances, let me know @yvespp 
* base branch needs to be changed after merging #653


yvespp:
@dowhiletrue the h2 db has to be updated here as well. Thanks!

yvespp (approved):

Issue #566, opened 2020-08-17
Replace net.sf.json-lib with another JSON library


yvespp:
net.sf.json-lib (net.sf.json imports) is very old and no longer maintained: https://mvnrepository.com/artifact/net.sf.json-lib/json-lib
We should replace it with a different JSON library. In some parts we already use GSON or maybe jax-b (`JsonbBuilder.create().toJson(object)`) can be used.

yvespp:
This apparently also pulls in common collections again (thanks @mburri):
```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ AMW_ear ---
[INFO] ch.puzzle.itc.mobiliar:AMW_ear:ear:1.17.17-SNAPSHOT
[INFO] \- ch.puzzle.itc.mobiliar:AMW_business:ejb:1.17.17-SNAPSHOT:compile
[INFO]    \- net.sf.json-lib:json-lib:jar:jdk15:2.4:compile
[INFO]       \- commons-collections:commons-collections:jar:3.2.1:compile
```

yvespp:
Fixed by https://github.com/liimaorg/liima/pull/669

PR #669, opened 2022-10-19
gh-566: replace net.sf.json.lib with jackson

dowhiletrue:


AMW_web/pom.xml
```diff
@@ -82,11 +82,10 @@
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-lang3</artifactId>
 		</dependency>
-		<dependency>
-			<groupId>net.sf.json-lib</groupId>
-			<artifactId>json-lib</artifactId>
-			<classifier>jdk15</classifier>
-		</dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
```
yvespp, line 86: Please fix indent and add `<scope>provided</scope>`

dowhiletrue: * indentation fixed (was not visible in my IDE, since this specific pom.xml uses tabs)
* `<scope>provided</scope>` is already inherited from `dependencyManagement`. added it anyway to avoid misunderstandings

yvespp (approved):

PR #685, opened 2023-01-09
547 update bootstrap js

mburri:
Upgrade bootstrap 3.x -> 5.x for AMW_Web

Replaces outdated library


yvespp:
@mburri I think bootstrap.min.js is not needed because bootstrap.bundle.min.js already includes it, can you remove it please?

yvespp:
I found some layout issues:
- The page content is no longer centered
- Releases are now sometimes multiline which breaks the table format
- The top menu (Apps, Servers, etc.) doesn't align with the right edge of the content anymore.
  image
- The content doesn't react to the browser window size correctly.
  image
- Servers page:
  - AppServer text field is not the same height as the other text fields on the page
- editResourceView:
  - Reset Label of a property is to close to the checkbox
    image

yvespp:
The *.map files probably shouldn't be included because of security reasons.

yvespp (approved):

PR #687, opened 2023-01-12
557 download allows content sniffing

StephGit:
Adds header to prevent content-sniffing in `angular` and `web`-module. 
(I guess the implementation in the web-module is obsolete and the Class `DeploymentCSVExport.java` is dead code.)


StephGit:
No not intentional. 
Done.

yvespp (approved):

Issue #682, opened 2023-01-03
Duplicate resource names via renaming


yvespp:
On resource creation via JSF UI the resource name check doesn't allow the same name with different case.
But on renaming a resource via UI (edit name property on the resource) the same name with different case is allowed.
Example:
* Create app: myApp
* Create app: MyApp => error
* Create app: MyApp1
* Edit MyApp1 and change name to MyApp => works, but shouldn't

Resource with the same name but different case shouldn't be allowed.

PR #706, opened 2023-10-16
fix: prevent renaming different resources to same name

llorentelemmc:
Issue #682 

Added validation to searchResourceByName() method so that renaming the same resource to uppercase or lowercase (eg, myapp to MyApp) is allowed, but not allowed for different resources.

### DIFF
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/resourcegroup/control/ResourcesScreenQueries.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/resourcegroup/control/ResourcesScreenQueries.java
@@ -39,9 +39,10 @@ public class ResourcesScreenQueries {
 				.setParameter("resourceTypeName", resourceTypeName);
 	}
 	
-	public Query searchResourceByName(String resourceName) {
-		return entityManager.createQuery("from ResourceEntity as res where res.name like :resourceName")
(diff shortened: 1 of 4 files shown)


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/resourcegroup/control/ResourcesScreenQueries.java
```diff
@@ -39,9 +39,10 @@ public Query searchResourceTypeByName(String resourceTypeName) {
 				.setParameter("resourceTypeName", resourceTypeName);
 	}
 	
-	public Query searchResourceByName(String resourceName) {
-		return entityManager.createQuery("from ResourceEntity as res where res.name like :resourceName")
-				.setParameter("resourceName", resourceName);
+	public Query searchResourceByName(String resourceName, Integer id) {
```
yvespp, line 42: Maybe this method should be renamed? `searchOtherResourcesWithName` or something?

yvespp (approved):

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/resourcegroup/control/ResourcesScreenQueries.java
llorentelemmc: Agreed - I have changed it

Issue #707, opened 2023-10-25
Removing relation of a consumed resource that has no relation name is not possible


llorentelemmc:
Setup: A resource with a related, consumed resource

Reproduce:
1. Remove the related resource

If you add a RelationName, the related resource is successfully removed. If there is no RelationName, the related resource is not removed.

The related resource should also be removable without relation name.

PR #709, opened 2023-10-26
fix: remove unnamed consumed relation

llorentelemmc:
Issue #707  

Simplified `doRemoveResourceRelationForAllReleases()` as it should not be necessary to identify a related resource as consumed or provided before deletion.


yvespp:
The loop over all the relations is needed actually. Resources can have multipile release (Release -> new in the UI).
If a resource with multiple releases is added to an other resource, then a relation is created between all resource releases.
So it will look like this:
```
r1_rl1 -> r2_rl1
       -> r2_rl2
```
So when a relation is deleted in the UI, all relations from r1_rl1 have to be deleted.

mburri:
Thanks for your input. I already suspected that we probably missed something.

What I also noticed: If you add a relation with multiple releases as a consumed resource and set the relation name - then the behavior is different: in this case only the currently selected relation is removed. Now we have the problem that removing unnamed resources does not work in our setup with wildfly.

I also noticed that there is zero test coverage and the code is quite hard to reason about. This means that the use cases (removing a relation from a resource) with all its special cases (resources with multiple releases, resourcetypes etc) are hard to understand and not documented anywhere.

I think it would be valuable to improve this situation:
- Document the different use cases for removing relationships from resources.
- Add tests to cover `ResourceRelationService#removeRelation(AbstractResourceRelationEntity relation)
- Code refactorings to improve code readability

Since issue #707 was not part of the current milestone, I'd recommend marking this PR as a draft and moving it to a later release.

yvespp:
Refactor, tests & docs is always good 👍

PR #711, opened 2023-11-02
fix: clipboard button

llorentelemmc:


AMW_angular/io/src/app/deployments/deployments.component.html
```diff
@@ -12,6 +12,11 @@
       [message]="errorMessage"
       [messageType]="'warning'"
     ></app-notification>
+    <div *ngIf="showToast" class="toast position-absolute top-0 end-0" role="alert" aria-live="assertive" aria-atomic="true">
```
mburri, line 15: This should probably go into a simple reusable component.

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -1,5 +1,5 @@
 import { Location } from '@angular/common';
-import { Component, OnInit, ViewChild } from '@angular/core';
+import {ChangeDetectorRef, Component, OnInit, ViewChild} from '@angular/core';
```
mburri, line 2: ChangeDetectorRef seems to be unused and should be removed

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -22,6 +22,13 @@ declare var $: any;
 @Component({
   selector: 'amw-deployments',
   templateUrl: './deployments.component.html',
+  styles: [`
```
mburri, line 25: If you create a dedicated toast component, this should go into this component too.

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -289,22 +296,22 @@ export class DeploymentsComponent implements OnInit {
     );
   }
 
-  copyURL() {
+  showToast: boolean = false;
```
mburri, line 299: Like in java classes, property are usually declared at the top and not in between methods. Just move it up to the top of the class

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -289,22 +296,22 @@ export class DeploymentsComponent implements OnInit {
...
-          '<input type="text" name="fname" class="textToCopyInput" style="opacity:0"/>'
-        ).val(url)
-      )
-      .find('.textToCopyInput')
-      .select();
     try {
-      document.execCommand('copy');
-    } catch (e) {
+      await navigator.clipboard.writeText(url);
+      console.log('URL copied to clipboard');
```
mburri, line 305: With the new toast and the window prompt if something fails, there is no need to log to the console.

AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -289,22 +296,22 @@ export class DeploymentsComponent implements OnInit {
...
+      this.showToast = true;
+      setTimeout(() => {
+        this.showToast = false;
+      }, 3000);
+
+    } catch (err) {
+      console.error('Failed to copy: ', err);
```
mburri, line 312: this console.log can be removed too.

mburri:
Seems good to me. @yvespp What do you think of the new Toast component that notifies the user that the URL was successfully copied to the clipboard?

yvespp:
Tested it and works. It's not very visible maybe it can be styled similar to the toast of the JSF app?
Example when a resource relation is removed:
image

llorentelemmc:
> Tested it and works. It's not very visible maybe it can be styled similar to the toast of the JSF app? Example when a resource relation is removed: image

Screenshot from 2023-11-06 17-06-08
@yvespp I've tried to style it more similarly to the notification of the JSF App, is it better like this?

yvespp:
@llorentelemmc would it be possible to add the blue part with the exclamation mark? If it's to complicated it's ok like this.

yvespp (approved):

llorentelemmc:
@yvespp I've added the blue part with the exclamation part and adjusted the colours a bit. What do you think?
Screenshot from 2023-11-09 17-20-23

yvespp:
Cool thanks!

Issue #353, opened 2018-04-23
One restcall to promote many deplyoments


andyzwa:
For promoting more than one deplyoment over the rest interface, it would be nicer to create all deployment in one rest call.
Actual for each deplyoment a (post)call have to be created.

yvespp:
Related to #440

mburri:
With the old frontend, it was possible to request multiple deployments at once:

![image](https://github.com/liimaorg/liima/assets/4577216/83352488-f468-4d0a-97f0-07ac6f231d04)

Note, that more than one deployment is checked in the first column.
When the button "redeploy" is clicked, a new screen allows to set parameters for the deployments:

![image](https://github.com/liimaorg/liima/assets/4577216/16e1ce44-9804-4e3f-a849-73cd90c12d84)

This allowed users to redeploy multiple applications and set the values (deployment date, environments, ...) only once.

With the new Frontend, redeployments are only possible for single deployments:

![image](https://github.com/liimaorg/liima/assets/4577216/f86973e4-7eab-4395-8d16-c31086131e6f)


If the user clicks "redeploy" the endpoint `POST /deployments` is called, implemented by `ch.mobi.itc.mobiliar.rest.deployments.DeploymentsRest#addDeployment`

We need to modify this endpoint (or add a new endpoint) that takes multiple applications and the configuration parameters to redeploy all applications at once. 
Then we need to change the front end to allow multiple deployments to be selected and redeployed.The selection is already there - but the redeploy action in the actions column only redeploys one application. I think we just need to add a "redeploy" button to the other buttons to initiate the workflow.

@yvespp are there other systems that use the current endpoint `POST /deployments` or is it only used by the angular GUI?

yvespp:
@mburri The post deployments endpoint is used by the maven plugin (https://github.com/liimaorg/liima-maven-plugin) as well as the go cli (https://github.com/liimaorg/liimactl/blob/master/cmd/deployment/create.go)
We use both to create deployments from our CI/CD tool.
I also found some calls where no user agent is set, not sure who creates them...

mburri:
@yvespp Thanks for your answer. In this case, we should probably add a new rest endpoint and not modify the existing one. 
I wonder if the new endpoint should allow both: single and multiple re-deployments.

yvespp:
Talked to @andyzwa , it's no longer relevant for them. So i will close this issue for now.

In case we reopen it:
[...]

PR #714, opened 2023-11-10
End-to-end tests for Liima

mburri:
I analyzed the existing implementation for deployments affecting #353. I quickly realized that there would be a huge potential for refactorings, but often the test coverage is not sufficient to support efficient refactorings. 

Writing new unit tests would be possible, but is often quite difficult due to the number of dependent services, repositories
and boundaries that must be mocked away. In addition, mocking all these dependencies actually makes it harder to refactor, since the mocks couple the test to the actual implementation.

Anyway. Having end-to-end tests that cover the main paths of the application would be helpful. There was an e2e test setup in the AMW_angular module. But:
* no tests have ever been written
* Protractor is deprecated as of September 2023: https://blog.angular.io/protractor-deprecation-update-august-2023-2beac7402ce0#:~:text=At%20the%20end%20of%20this,change%20in%20August%20of%202022.

So I removed Protractor from the AMW_angular module.

Then I decided to create a separate module for the end-to-end testing, because the whole application is affected, not just the parts implemented in Angular.

For now, this is just the setup with a very simple test to show, that the setup is correct. 

This PR adds a dedicated Github Action Workflow to run the end-to-end tests. But there are lots of overlappings with the existing build and test action. The e2e tests could also be added to the existing workflow.

Please feel free to suggest changes, improvements or even reject this PR


yvespp:
Im fine with integrating it into the existing workflow, I think we want to run it always anyway.
I'm not familiar with cypress but it looks good. We can also use it to test REST APIs that aren't used by Angular.

yvespp (approved):

PR #715, opened 2023-11-15
Update Angular to v17

llorentelemmc:
Updated Angular to v17, including any fixes to breaking changes


mburri:
I pushed two new commits:
1. Update the node version in `.nvmrc` the v18 (lts/hydrogen)
2. Using 127.0.0.1 instead of localhost resolves the econnrefused problem. Node 18 does not prefer ipv4 over ipv6 anymore. This means that its not garantueed that localhost resolved. See https://github.com/chimurai/http-proxy-middleware#nodejs-17-econnrefused-issue-with-ipv6-and-localhost-705

yvespp (approved):

PR #717, opened 2023-11-22
Migration to angular standalone

mburri:
This PR contains the migration to the new angular standalone components.


mburri:
I'm closing the pull request. I think its better to migrate the create tags page first (and merge it to main) - and then migrate to standalone components to separate new features from the migrations

PR #719, opened 2023-12-07
Settings -> Tags Angular Migration

mburri:
This PR includes:
* Settings -> Tags in migrated to angular
* New endpoints to list, create and delete (global) tags
* the roles/ permissions page was moved to the settings module
* styles for roles/ permission where adjusted to look like /settings/tags

This PR does not include:
* Deletion of JSF Pages and Beans for settings/ tags. This will be a separate PR
* Cleanups in the existing angular code - there is a huge potential to clean up lots of code. Code cleanups will be done with future development

A lot of the changes in the files are caused by new formatting and linting rules. We now use prettier consistently which causes lots of reformattings


yvespp (approved):

mburri:
@yvespp I've just pushed a change to make the screens a bit wider - they still don't take up 100% of the screen for wide displays.

PR #721, opened 2023-12-14
fix error when sending mail for archived deployment, fix sending mail on predeploy failure

yvespp:


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/control/DeploymentNotificationService.java
```diff
@@ -55,44 +52,35 @@ public class DeploymentNotificationService {
...
-						getMessageContentForDeployments(deployments),
-						emailReceipients)) {
-					message = getSuccessfulSendMailToReceipientMessage(emailReceipients);
-				}
-			} catch (MessagingException e) {
-				message = getFailureSendMailMessage(e.getMessage());
-				log.log(Level.WARNING, "Deployment notification ("
-						+ subjectMessage + ") could not be sent", e);
+		if (deployments == null || deployments.isEmpty()) {
+			return message;
```
mburri, line 72: So - if deployments is null or empty, null will be returned? is this intentional?

mburri (approved):

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/control/DeploymentNotificationService.java
yvespp: I think so. It was returning null before, I just switched the if statement around.

PR #726, opened 2024-01-03
Feat/725 replace application info page

StephGit:


AMW_angular/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
yvespp, line 1: This file is probably not needed

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
yvespp, line 1: This file is probably not needed

AMW_angular/io/src/app/settings/settings.component.html
```diff
@@ -12,7 +12,7 @@
         <a class="nav-link" href="/AMW_web/pages/settings.xhtml?mode=STP">STP Management</a>
         <a class="nav-link" href="/AMW_web/pages/settings.xhtml?mode=releases">Releases</a>
         <a class="nav-link" routerLink="/settings/permission" routerLinkActive="active">Roles and Permissions</a>
-        <a class="nav-link" href="/AMW_web/pages/settings.xhtml?mode=appVersion">Application Info</a>
+        <a class="nav-link" routerLink="/settings/application-info" routerLinkActive="active">Application Info</a>
       </nav>
     </div>
     <div class="col-9">
```
yvespp, line 18: Can we go to `col-10` here? With `cole-9` the page has unused extra space on the right:
image
This page could benefit from some more space and this size is also in line with the other pages (deploy and audit).

yvespp (commented):
Cool, looks good otherwise!

AMW_angular/io/src/app/settings/settings.component.html
StephGit: yes, makes sense

yvespp (approved):

PR #735, opened 2024-01-25
Angular 17.1, faster builds with vite and reduced bundle sizes

mburri:
This MR contains the following changes:
* Upgrade to Angular 17.1
* Switch to Angulars `browser-esbuild` for improved compile/ build performance when developing locally
* Enable bundle opimizations for frontend builds
* Enable gzip compression for frontend assets in wildfly

The changes drastically improve the applications lighthouse score

image


yvespp:
Could we switch to the `application` builder in the future? Would that be a lot of work?

PR #736, opened 2024-01-30
Feature/413 limit deployment log file size

### LINKED ISSUE #413: Limit deployment log file size to 10MB
We had a case where a deployment script wrote several GBs to stdout. If you try to open this log file in the GUI you get an OutOfMemory exception. By limiting the log file size this should be prevented.

mburri:
To be merged after #735 

fixes #413


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentLog.java
yvespp, line 1: Can the three classes `DeploymentLog`, `DeploymentLogContent` and `DeploymentLogContent` be merged into one class? Why does the id type differ per class (int, Integer and long)?

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentBoundary.java
```diff
@@ -801,16 +801,19 @@ public String[] getLogFileNames(final int deploymentId) {
...
+    public String getDeploymentLog(String logName) throws IOException {
         String logsPath = ConfigurationService.getProperty(ConfigKey.LOGS_PATH);
-        if (logName.contains(File.separator)) {
-            throw new IllegalAccessException("The log file contains a file separator (\"" + File.separator + "\"). For security reasons, this is not permitted!");
-        }
+
+        String name = logsPath + File.separator + logName;
+        File file = new File(name);
+
+        if (file.length() > 1_000_000) throw new IOException(String.format("%s is larger than 10 MB and cannot be shown.", file.getName()));
```
yvespp, line 810: Instead of an exception, could we instead read the first 10MB of the file and append an error to it?

yvespp: We could also try to stream the file instead to prevent loading it into memory completely.

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentLog.java
mburri: Hmm... there seems to be only one `DeploymentLogContent` - did you mean `DeploymentLogContentCommand`

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentBoundary.java
mburri: Both are good suggestions - I'll have to check it. I wonder if the users browser will crash if we stream large files.

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentLog.java
mburri: I merged `DeploymentLog` and `DeploymentLogContent` and cleaned up the various types.
`DeploymentLogContentCommand` is used to encapsulate the input validation.

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/deploy/boundary/DeploymentBoundary.java
mburri: I've chosen a third option: 

Only the last 10mb of the log file is shown (if larger) and a message is prepended that says that some part of the log is missing...

I imagine that the interesting log entries are at the end of the log - and not at the beginning. 

@yvespp What do you think about it?

yvespp (approved):

PR #739, opened 2024-02-06
feature: migrate deployment parameter page to Angular and update to version 17.1

llorentelemmc:
@yvespp while adding user permissions to deployment-parameter and tags components in Angular, I noticed that permissions are defined as either old or new (`AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/security/entity/Permission.java`). Do you know the reason why some are defined as old and some as new, and is it okay that `MANAGE_GLOBAL_TAGS` and `MANAGE_DEPLOYMENT_PARAMETER` are old?


AMW_web/src/main/webapp/pages/settings.xhtml
StephGit, line 1: The component reference at line 180 can be removed, same for the component `deploymentParameterComp.xhtml` and it's view controller `DeploymentParameterView.java`

AMW_angular/io/src/app/settings/deployment-parameter/deployment-parameter.component.html
```diff
@@ -0,0 +1,41 @@
...
+    <thead>
+    <tr>
+      <th>Name</th>
+      <th></th>
+    </tr>
+    </thead>
+    <tbody>
+    @for (paramKey of paramKeys; track paramKey) {
+    <tr>
+      <td>{{ paramKey.key }}</td>
```
StephGit, line 29: Name and icon should be on the same line.

AMW_angular/io/src/app/settings/deployment-parameter/deployment-parameter.component.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [DeploymentParameterComponent]
+    })
+    .compileComponents();
+    
+    fixture = TestBed.createComponent(DeploymentParameterComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
```
StephGit, line 17: Add missing HttpClient -> see failing test

AMW_angular/io/src/app/settings/deployment-parameter/deployment-parameter.component.ts
```diff
@@ -0,0 +1,61 @@
...
+@Component({
+  selector: 'app-deployment-parameter',
+  standalone: true,
+  imports: [CommonModule, ReactiveFormsModule, ToastComponent, FormsModule],
+  templateUrl: './deployment-parameter.component.html',
+  styleUrl: './deployment-parameter.component.scss',
+})
+export class DeploymentParameterComponent {
+  keyName = '';
+  paramKeys: Key[] = [];
```
StephGit, line 18: Consider using a observable and asyncPipe for showing new entries in the list.

1. replace the Array with an Observalbe<Key[]>
2. create a method which loads all entries (hint: same like the get request but instead of subscribe use pipe()
3. call the method from constructor
4. call the method on the successful post
5. call the method on the successful delete

AMW_angular/io/src/app/settings/deployment-parameter/deployment-parameter.component.html
```diff
@@ -0,0 +1,41 @@
...
+        <td>
+          <button class="btn" (click)="deleteKey(paramKey.key)" [attr.data-cy]="'delete-' + paramKey.key">
+            <i class="bi bi-trash"></i>
```
StephGit, line 32: The button should be styled and the IconComponent can be used for a nice styling of the icon.
similar example: https://github.com/liimaorg/liima/blob/6549fa510efe5c5245a2abe9b8466f039f7a8126/AMW_angular/io/src/app/settings/releases/releases.component.html#L52

AMW_angular/io/src/app/settings/deployment-parameter/deployment-parameter.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  @ViewChild(ToastComponent) toast: ToastComponent;
+
+  constructor(private http: HttpClient) {
+    http.get<Key[]>('/AMW_rest/resources/deployments/deploymentParameterKeys/').subscribe({
```
StephGit, line 23: `/ ` there is no need for the backslash - same for post

AMW_angular/io/src/app/settings/deployment-parameter/deployment-parameter.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  addKey(): void {
+    const trimmedKeyName = this.keyName.trim();
+    if (trimmedKeyName.length > 0 && trimmedKeyName.toLowerCase() !== 'null') {
+      this.http
+        .post<Key>('/AMW_rest/resources/deployments/deploymentParameterKeys/', { key: this.keyName, value: null })
+        .subscribe({
+          next: (newKey) => {
+            this.paramKeys.push(newKey);
```
StephGit, line 37: The new entry will no be displayed. This is because of the response type which will not map to a Key-entity. You have to create a new Key-object first and then push it to the array.

(Hint: check the response value in the network tab)

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
```diff
@@ -363,6 +367,32 @@ public Response getAllDeploymentParameterKeys() {
         return Response.status(Status.OK).entity(deploymentParameters).build();
     }
 
+    @POST
+    @Path("/deploymentParameterKeys/")
```
StephGit, line 371: Remove backslash ( as well on line 360)

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
```diff
@@ -363,6 +367,32 @@ public Response getAllDeploymentParameterKeys() {
...
+    @DELETE
+    @Path("/deploymentParameterKeys/{keyName}")
+    @ApiOperation(value = "Deletes one parameter key")
+    public Response deleteOneKey(@PathParam("keyName") String keyName) {
+        Key keyToBeRemoved = keyRepository.findFirstKeyByName(keyName);
+        if (keyToBeRemoved != null) {
+            deploymentParameterBoundary.deleteDeployParameterKey(keyToBeRemoved);
+            return Response.ok().build();
+        } else {
+            return Response.status(Response.Status.NOT_FOUND).build();
```
StephGit, line 392: throw NotFoundException instead of  if-else. So the endpoint only returns ok-responses. The error will be mapped in the service by using `NotFoundExceptionMapper`

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
```diff
@@ -363,6 +367,32 @@ public Response getAllDeploymentParameterKeys() {
         return Response.status(Status.OK).entity(deploymentParameters).build();
     }
 
+    @POST
+    @Path("/deploymentParameterKeys/")
+    @ApiOperation(value = "Adds one parameter key")
+    public Response addOneParameterKey(DeploymentParameterDTO deploymentParameterDTO) {
+        Key keyToBeAdded = new Key(deploymentParameterDTO.getKey());
+        keyRepository.createDeployParameterKey(keyToBeAdded);
```
StephGit, line 375: Direct access of the repository is not recommended. Please use the according function in the boundary package

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/deployments/DeploymentsRest.java
```diff
@@ -363,6 +367,32 @@ public Response getAllDeploymentParameterKeys() {
...
+                .entity(keyToBeAdded)
+                .location(URI.create("/deploymentParameterKeys/" + keyToBeAdded.getName()))
+                .build();
+    }
+
+    @DELETE
+    @Path("/deploymentParameterKeys/{keyName}")
+    @ApiOperation(value = "Deletes one parameter key")
+    public Response deleteOneKey(@PathParam("keyName") String keyName) {
+        Key keyToBeRemoved = keyRepository.findFirstKeyByName(keyName);
```
StephGit, line 387: same like above

AMW_angular/io/src/app/settings/tags/tags.component.html
```diff
@@ -8,27 +8,26 @@ <h1>Tags</h1>
...
       </tr>
     </thead>
     <tbody>
       @for (tag of tags; track tag) {
       <tr>
         <td>{{ tag.name }}</td>
         <td>
-          <button class="btn" (click)="deleteTag(tag.id)" [attr.data-cy]="'delete-' + tag.name">
+          <button class="btn btn-sm btn-danger me-0" (click)="deleteTag(tag.id)" [attr.data-cy]="'delete-' + tag.name">
             <i class="bi bi-trash"></i>
```
StephGit, line 31: use icon-component

StephGit (approved):

yvespp:
When I try to add a duplicate tag, the error toast sometimes shows for a short time only (less than a second, sometime it just blinks). The only reliable way I found to reproduce this spamming the add button with the same tag.

yvespp (approved):
It looks good to me now, thanks!
The toast still has a strange behaviour. It looks like the timer on the component keeps running even if a new toast is show (replacing the old before it's time) and then the old timer closes the new toast. Timer: https://github.com/liimaorg/liima/blob/7c4fa312f2d3ce29161a81882305e484ce4827e1/AMW_angular/io/src/app/shared/elements/toast/toast.component.ts#L45
May the old timer has to be canceled or it should only hide the toast if it matches the toast.
Should we merge this and you have a look in a separate issue?

StephGit:
We resolved the timeout issue with clearing the old one first. 

A improvement for the future would be, to show multiple toast instead of only one. Should we create a story for a toastContainer-Component?

StephGit (approved):

yvespp:
> A improvement for the future would be, to show multiple toast instead of only one. Should we create a story for a toastContainer-Component?

I created an new issue for it: https://github.com/liimaorg/liima/issues/744
NG-Bootstrap already hase something like this, maybe that could be used?

PR #764, opened 2024-08-23
refactor: curry isAllowed to clean up callers

mburri:


StephGit:
Thanks for your contribution @mburri 

I really like the functional approach, but for developers who are not familiar with functional programming, it's harder to understand.

Could you maybe add some documentation

Issue #740, opened 2024-02-06
Replace JSF GUI in Settings -> Property Types page with Angular


llorentelemmc:
The AMW GUI should further be replaced with Angular: https://github.com/liimaorg/liima/issues/16
Next page we should replace is Settings - Property Types

yvespp:
Fixed by https://github.com/liimaorg/liima/pull/739

PR #765, opened 2024-08-30
Feature/740 angular property types

StephGit:
#740

Refactor settings-> property types


yvespp:
When to user only has read permissions, the columns Edit and Delete should be hidden and the button `Add property type` shouldn't be shown.

yvespp (approved):

StephGit:
Fixed the permissions based on https://github.com/liimaorg/liima/pull/768.

The permissions `EDIT_PROP_TYPE_NAME`, `EDIT_PROP_TYPE_VALIDATION` and `SAVE_SETTINGS_PROPTYPE` are now handled equal. Is there a need for specific handling, for example, a user can only update the property-type-name not it's validation?

PR #766, opened 2024-09-02
Remove My favorites

llorentelemmc:
To reduce complexity of the codebase, unused funtionality is being removed. Here we are removing My favorites and anything related to it.


yvespp:
The DB can also be cleaned up via Liquibase script.

PR #781, opened 2024-10-16
feat: add environments page with new ui without functionality

BeriBoss:


AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/environments/EnvironmentsRest.java
```diff
@@ -42,4 +43,13 @@ public List<EnvironmentDTO> getEnvironments(@ApiParam("Also returns Environment
         return environments;
     }
 
+    @GET
+    @Path("/contexts")
+    @ApiOperation(value = "Get all contexts", notes = "Returns all contexts as environments")
+    public List<EnvironmentDTO> getContexts() {
+        List<EnvironmentDTO> environments = new ArrayList<>();
+        contextLocator.getAllEnvironments().stream().map(EnvironmentDTO::new).forEach(environments::add);
+        return environments;
```
mburri, line 52: You don't need the `forEach` to get from the stream to the list. You can use `stream().collect(Collectors.toList())`

Your implementation would look like this:
```
public List<EnvironmentDTO> getContexts() {
    return contextLocator.getAllEnvironments().stream().map(EnvironmentDTO::new).collect(Collectors.toList());
}
```

This eliminates the mutation (bad) of the temporary variable (code smell!) `environments`..

AMW_angular/io/src/app/settings/environments/environments-page.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  private environmentsService = inject(EnvironmentService);
+  private toastService = inject(ToastService);
+  private error$ = new BehaviorSubject<string>('');
+  private destroy$ = new Subject<void>();
+  environmentTree: EnvironmentTree[];
+
+  ngOnInit(): void {
+    this.error$.pipe(takeUntil(this.destroy$)).subscribe((msg) => {
+      msg !== '' ? this.toastService.error(msg) : null;
+    });
```
mburri, line 25: I think you don't need to handle errors here. Errors from http requests are already handled globally.

And If you need to show an error from within your component, just use the toastservice (or maybe throw an error)

AMW_angular/io/src/app/settings/environments/environments-page.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    //this.getUserPermissions(); //TODO Handle Permissons
+    this.environmentsService.getContexts().subscribe((contexts) => {
+      this.environmentTree = this.buildEnvironmentTree(contexts)[0].children;
+    });
+  }
```
mburri, line 30: It would be nice to convert `this.environmehtTree` to a `computed` signal.

first (not in ngOnInit) but as class property, define a signal for your contexts (the response from the server) and then use the computed funtion to get a signal for your environmnetTree:

```
contexts = toSignal(this.environmentsService.getContexts())

environmentTree = computed(() => 
    if(!this.contexts()) return // guard: wait until environments are loaded
    return this.buildEnvironmentTree(this.contexts())[0].children;
)

File: AMW_angular/io/src/app/settings/environments/environments-page.component.ts [hunk 4]
```diff
@@ -0,0 +1,48 @@
...
+          new EnvironmentTree(
+            environment.id,
+            environment.name,
+            environment.nameAlias,
+            this.buildEnvironmentTree(environments, environment.name),
+            environment.selected,
+            environment.disabled,
+          ),
+      );
+    return envTree;
```
mburri, line 46: Remove the temporary variable - just return what you get from streaming over the environments:

```
return environments
      .filter((environment) => environment.parent === parentName) // Find items with the current parentId
      .map(
        (environment) =>
          new EnvironmentTree(
            environment.id,
            environment.name,
            environment.nameAlias,
            this.buildEnvironmentTree(environments, environment.name),
            environment.selected,
            environment.disabled,
          ),
      );

File: AMW_angular/io/src/app/settings/environments/environments-page.component.ts [hunk 5]
```diff
@@ -0,0 +1,48 @@
...
+import { ToastService } from '../../shared/elements/toast/toast.service';
+import { BehaviorSubject, Subject } from 'rxjs';
+import { Environment, EnvironmentTree } from '../../deployment/environment';
+import { takeUntil } from 'rxjs/operators';
+
+@Component({
+  selector: 'app-environments-page',
+  standalone: true,
+  imports: [IconComponent],
+  templateUrl: './environments-page.component.html',
```
mburri, line 13: Do not forget to add `ChangeDetectionStrategy.ON_PUSH`

AMW_angular/io/src/app/deployment/environment.ts
```diff
@@ -6,3 +6,28 @@ export interface Environment {
   selected: boolean;
   disabled: boolean;
 }
+
+export class EnvironmentTree {
```
mburri, line 10: Personally - I don't like classes.

I'd probably just define a type and thats it. If it matches the Response from the backend, you won't need a constructor. If necessary, you could create a simple function that converts between the API and your type.

AMW_angular/io/src/app/settings/environments/environments-page.component.ts
mburri: if you just defined the type (see not to class EnvironmentTree), it gets even shorter:

```
return environments
      .filter((environment) => environment.parent === parentName) // Find items with the current parentId
      .map(
           (environment) =>
                { return
                     {...environment, 
                     children: this.buildEnvironmentTree(environments, environment.name),
                    }
                 }
      );

mburri (changes requested):

mburri (approved):

PR #784, opened 2024-10-28
Feature/796: add functionality to environments page with new ui

BeriBoss:
Add **functionality** to environments page with new ui:
- add new environment/domain
- update existing environment/domain
- delete environment/domain


AMW_angular/io/src/app/settings/environments/environment-delete/environment-delete.component.ts
```diff
@@ -0,0 +1,31 @@
+import { Component, EventEmitter, inject, Input, Output } from '@angular/core';
+import { NgbActiveModal } from '@ng-bootstrap/ng-bootstrap';
+import { Environment } from '../../../deployment/environment';
+import { ModalHeaderComponent } from '../../../shared/modal-header/modal-header.component';
+
+@Component({
+  selector: 'app-function-delete',
```
mburri, line 7: this should probably be `app-environment-delete`

AMW_angular/io/src/app/deployment/environment.service.ts
```diff
@@ -39,4 +48,40 @@ export class EnvironmentService extends BaseService {
       .get<Environment[]>(`${this.getBaseUrl()}/environments/contexts`)
       .pipe(catchError(this.handleError));
   }
+
+  save(environment: Environment) {
+    if (environment.id) {
+      return this.update(environment);
+    } else {
+      return this.create(environment);
+    }
```
mburri, line 57: more concise would be: 
```
save(environment: Environment) {
    if (environment.id) return this.update(environment);
    return this.create(environment)
}

File: AMW_angular/io/src/app/settings/environments/environments-page.component.html [hunk 3]
```diff
@@ -1,48 +1,48 @@
 <div class="container w-50">
   <h1>Environments</h1>
-  <button type="button" class="btn btn-primary">
+  <button type="button" class="btn btn-primary" (click)="addDomain()">
     <app-icon icon="plus-circle"></app-icon>
     Add Domain
   </button>
-  @for(env of environmentTree(); track env.id){
+  @for(dom of environmentTree(); track dom.id){
```
mburri, line 7: is `dom` shorthand for `domain`? 
My first association was `Document Object Model` - personally, I'd prefer `domain` instead of `dom`

AMW_angular/io/src/app/settings/environments/environments-page.component.ts
```diff
@@ -2,36 +2,142 @@ import { ChangeDetectionStrategy, Component, computed, inject, Signal } from '@a
...
   templateUrl: './environments-page.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class EnvironmentsPageComponent {
   private environmentsService = inject(EnvironmentService);
-  //TODO Handle permissions
-  contexts: Signal<Environment[]> = toSignal(this.environmentsService.getContexts());
+  private modalService = inject(NgbModal);
+  private toastService = inject(ToastService);
+  //TODO handle permissions
```
mburri, line 24: Implement or remove TODO

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/environments/EnvironmentsRest.java
```diff
@@ -24,10 +28,13 @@ public class EnvironmentsRest {
     @Inject
     ContextLocator contextLocator;
 
+    //TODO: EnvironmentsScreenDomainService is deprecated.
```
mburri, line 31: remove the todo: the class is already annotated with `@deprecated` - the TODO does not help here.

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/environments/EnvironmentsRest.java
```diff
@@ -50,4 +57,36 @@ public List<EnvironmentDTO> getContexts() {
...
+    }
+
+    @PUT
+    @Path("/contexts/{id : \\d+}")
+    @ApiOperation(value = "Update existing context")
+    public Response updateContext(@ApiParam("Environment ID") @PathParam("id") Integer id, EnvironmentDTO request) {
+        try {
+            environmentsScreenDomainService.saveEnvironment(id, request.getName(), request.getNameAlias());
+            return Response.status(Response.Status.OK).build();
+        } catch (Exception e) {
```
mburri, line 80: do not catch all exceptions here.
environmentScreenDomainService should only throw `ResourceNotFoundExceptions` 

This should already be covered by `NoResultExceptionMapper`

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/environments/EnvironmentsRest.java
```diff
@@ -50,4 +57,36 @@ public List<EnvironmentDTO> getContexts() {
...
     }
 
+    @POST
+    @Path("/contexts")
+    @ApiOperation(value = "Add new context")
+    public Response addContext(@ApiParam() EnvironmentDTO request) {
+        try {
+            environmentsScreenDomainService.createContextByName(request.getName(), request.getParentId());
+            return Response.status(Response.Status.OK).build();
+        } catch (Exception e) {
```
mburri, line 67: see comment below

mburri (changes requested):

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/exceptions/ElementAlreadyExistsExceptionMapper.java
```diff
@@ -0,0 +1,16 @@
...
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+@Provider
+public class ElementAlreadyExistsExceptionMapper implements ExceptionMapper<ElementAlreadyExistsException> {
+    @Override
+    public Response toResponse(ElementAlreadyExistsException exception) {
+        return Response.status(Response.Status.INTERNAL_SERVER_ERROR).entity(new ExceptionDto(exception)).build();
```
mburri, line 13: Is there a better option than INTERNAL_SERVER_ERROR? The use won't now what went wrong! Maybe add an error message?

Probably one of these: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

Maybe 409 Conflict

mburri (changes requested):

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/exceptions/ElementAlreadyExistsExceptionMapper.java
BeriBoss: Do I really need to add an error message? If so, what is the best way to do it? Because I think wherever this exception is thrown, a specific message is already set when the exception is created? Like 

```
if (getContextByName(newName) != null) {
    String message = "Der Kontext mit dem Namen " + newName + " ist bereits vorhanden und kann nicht erstellen werden";
   throw new ElementAlreadyExistsException(message, ContextEntity.class, newName);
            }
```
And when I create an ExceptionDTO with a Throwable ( `new ExceptionDto(exception)` ), the message of the Throwable is set in the ExceptionDTO:
```
public ExceptionDto(Throwable throwable) {		
	this.message = throwable.getMessage();
}
```

 So I think this particular message is better as if I would return some message. Or not?

mburri: :+1: 
ok - no need for a specific message

PR #785, opened 2024-10-28
Button component

llorentelemmc:
A reusable button component

### DIFF
--- a/AMW_angular/io/src/app/shared/button/button.component.ts
+++ b/AMW_angular/io/src/app/shared/button/button.component.ts
@@ -1,10 +1,15 @@
-import { Component, input, output } from '@angular/core';
+import { Component, input } from '@angular/core';
 import { NgClass } from '@angular/common';
 
 @Component({
   selector: 'app-button',
   template: `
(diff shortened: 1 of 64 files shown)


AMW_angular/io/src/app/shared/button/button.component.ts
```diff
@@ -0,0 +1,19 @@
...
+    <button type="button" class="btn" [ngClass]="class()" [disabled]="disabled()" (click)="clicked.emit()">
+      <ng-content select="[data-cy-slot]"></ng-content>
+      <ng-content></ng-content>
+    </button>
+  `,
+  standalone: true,
+  imports: [NgClass],
+})
+export class ButtonComponent {
+  class = input<string>('');
```
mburri, line 16: maybe create a "variant" property that could be "danger", "primary" and the apply the style with `class="'btn btn-' + variant()"`
Otherwise, the caller needs to specify the whole class "btn-primary" - and its not type safe

mburri (changes requested):

mburri (approved):

PR #791, opened 2024-11-07
Feature/1022 codemirror v6

StephGit:
Adds code- and diff-editor-component
Fullscreen-Toggle (usefull for diff-view)
Refactor to use new app components
Fix and extend tests


AMW_angular/io/src/app/settings/functions/function-edit.component.html
```diff
@@ -5,38 +5,43 @@
...
+      }
     </div>
-    <div class="mb-3" *ngIf="revisions.length > 0">
+    @if (revisions.length > 0) {
+    <div class="mb-3">
       <label class="form-label">Compare with history</label>
       <div ngbDropdown>
-        <!-- TODO: convert this button-->
-        <button class="btn btn-secondary dropdown-toggle" id="revisionDropdown" ngbDropdownToggle>
+        <app-button [variant]="'secondary'" ngbDropdownToggle [dataCy]="'button-drpdown'">
```
mburri, line 26: drpdown -> dropdown?

AMW_angular/io/src/styles.scss
```diff
@@ -15,8 +15,10 @@ $toast-padding-y: 0;
 
 @import 'bootstrap/scss/bootstrap';
 @import '@ng-select/ng-select/themes/default.theme.css';
+/*
 @import 'codemirror/lib/codemirror';
 @import 'codemirror/addon/dialog/dialog.css';
+*/
```
mburri, line 21: delete unused code?

mburri (changes requested):
I think that the editor should be higher than only one line when you create a new function
image

on the other hand, there should be probably an upper limit for the height?

image

StephGit:
> I think that the editor should be higher than only one line when you create a new function
> on the other hand, there should be probably an upper limit for the height?

made a fix for overflow/ height limit. 
quick fix with initial container size

llorentelemmc:
Screenshot from 2024-11-08 12-54-45
Screenshot from 2024-11-08 12-56-19
In the new editor it's not immediately obvious which side is the current version.
Additionally, in the old editor only the current version was editable in the merge view.

AMW_e2e/cypress/e2e/settings/functions/crud.cy.js
```diff
@@ -40,4 +40,32 @@ describe("Functions -CRUD", () => {
     cy.get('[data-cy="button-delete"]').click();
     cy.contains("Function deleted");
   });
+
+  it("should create, edit, compare and delete a function", () => {
```
llorentelemmc, line 44: You could also describe the new test scenario at the top of the file?

llorentelemmc:
Screenshot from 2024-11-11 10-35-12
Screenshot from 2024-11-11 10-31-08
The revisions list is hidden when fullscreen toggle is selected

StephGit:
Rearranged ui to improve visibility of dropdown

PR #793, opened 2024-11-08
Feature/1004 servers implement filter search

BeriBoss:
Please be aware that this branch [feature/937-servers-list](https://github.com/liimaorg/liima/tree/feature/937-servers-list) must first be merged into development.


BeriBoss:
As I have seen in the old code and if I have understood it correctly, I do not need to add any further permissions in the filter, is that right? @StephGit

AMW_angular/io/src/app/servers/servers-filter/servers-filter.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  runtimes = input.required<Resource[]>();
+  appServerSuggestions = input.required<string[]>();
+  searchFilter = output<ServerFilter>();
+  selectedEnvironmentName: string = 'All';
+  selectedRuntimeName: string = 'All';
+  appServer: string;
+  host: string;
+  node: string;
+
+  search: OperatorFunction<string, readonly string[]> = (text$: Observable<string>) =>
```
BeriBoss, line 29: I have copied and pasted this one (https://ng-bootstrap.github.io/#/components/typeahead/examples), can we have a look at it together? @StephGit

BeriBoss: I will keep it that way. Thank you!

AMW_angular/io/src/app/servers/servers-filter/servers-filter.component.html
```diff
@@ -0,0 +1,69 @@
...
+                <div class="col-sm-1">
+                    <label for="node" class="form-label">Node</label>
+                </div>
+                <div class="col-sm-3">
+                    <input type="text" class="form-control" id="node" [(ngModel)]="node">
+                </div>
+            </div>
+        </div>
+        <div class="card-footer">
+            <div class="row justify-content-between">
```
StephGit, line 61: Buttons could be align next to each other on the right? Would be similar to apps-page and modals

AMW_angular/io/src/app/servers/servers-filter/servers-filter.component.html
```diff
@@ -0,0 +1,69 @@
+<div class="container">
+    <div class="card">
```
StephGit, line 2: Instead of returning a `card` the filter component should only return a container. similar to apps-filter.component.html
The card should be used for both components in server -> #1032

StephGit: This should be resolved with #1032

StephGit (approved):

PR #799, opened 2024-11-15
feat: change routing from settings

BeriBoss:


StephGit (approved):
All old ui-components should be deleted.

PR #818, opened 2024-12-24
Feat/list functions 1040

StephGit:


AMW_angular/io/src/app/resources/resource-edit/resource-edit.component.html
```diff
@@ -16,17 +16,16 @@
       Loaded resource:
```
llorentelemmc, line 16: Remove this part? Resource name already visible in page-title

PR #827, opened 2025-01-13
Angular: make tables more similar, fix some UI bugs

yvespp:
Can you please check if my changes are ok?I tried to make all pages look the same but there are still some open issues.

Can you please create all future tables the same look please?

### Settings

- [x] Opening the settings page opens the environments page, but it isn't selected in the menu to the right.
- [x] When opening a settings page, tabt gets renamed to Settings and stays that way when switching to other pages. I would be nice it every page would write to the table title in the following schema: <page title> - Liima
- [x] There is an error when saving a new function:
```
Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: button
```
- [x] When creating or changing a release, the saved and displayed release will always be one day before the date selected. E.g. entring 04.01.2025 will be saved as 03.01.2025.

### Apps page
- [x] Somehow the table has a lighter gray for the table striping than the other pages. Also the table header is colored but not on the other pages.
- [x] With "Add Application" all resources can be select instead of just AppServers, same happens on the deployment page.
- [x] When saving, the loading spinner doesn't disappear and the page has to be reloaded. 
- [x] Can the table filters please be added to the url, so the page can be bookmarked?
- [x] If an AppServer or App is created with a name of any existing resource, an empty toast error is shown.
- [x] Can you please decrease the margines between the apps and the appServers in the table, so it gets more compact.
- [x] Searching for an appServer/app doesn't reload the pagination and shows no result when clicked.

### Server page
- [x] Can the table filters please be added to the url, so the page can be bookmarked?
- [x] Same as on the Apps page, the table header is gray and the table stiping is lighter then the on the settings page.

### Resources page
- [x] Could resource types that have resource not be indente but instead align the names and shift the + sign to the left?

Please push commits to this branch to fix the open issues and the check them off.

### DIFF
--- a/AMW_angular/io/src/app/auditview/auditview.routes.ts
+++ b/AMW_angular/io/src/app/auditview/auditview.routes.ts
@@ -1,6 +1,6 @@
 import { AuditviewComponent } from './auditview.component';
 
 export const auditviewRoutes = [
-  { path: 'auditview', component: AuditviewComponent },
-  { path: 'auditview/:resourceId', component: AuditviewComponent },
+  { path: 'auditview', component: AuditviewComponent, title: 'Audit View - Liima' },
+  { path: 'auditview/:resourceId', component: AuditviewComponent, title: 'Audit View - Liima' },
 ];
--- a/AMW_angular/io/src/app/auditview/auditview.routes.ts
+++ b/AMW_angular/io/src/app/auditview/auditview.routes.ts
@@ -1,6 +1,7 @@
 import { AuditviewComponent } from './auditview.component';
 
+const AUDIT_VIEW_TITLE = 'Audit View - Liima';
 export const auditviewRoutes = [
-  { path: 'auditview', component: AuditviewComponent, title: 'Audit View - Liima' },
-  { path: 'auditview/:resourceId', component: AuditviewComponent, title: 'Audit View - Liima' },
+  { path: 'auditview', component: AuditviewComponent, title: AUDIT_VIEW_TITLE },
+  { path: 'auditview/:resourceId', component: AuditviewComponent, title: AUDIT_VIEW_TITLE },
 ];
--- a/AMW_angular/io/src/app/deployment/deployment-routes.ts
+++ b/AMW_angular/io/src/app/deployment/deployment-routes.ts
@@ -1,10 +1,11 @@
 import { DeploymentComponent } from './deployment.component';
 
 export const deploymentRoutes = [
-  { path: 'deployment', component: DeploymentComponent },
-  { path: 'deployment/:deploymentId', component: DeploymentComponent },
+  { path: 'deployment', component: DeploymentComponent, title: 'Deployment - Liima' },
+  { path: 'deployment/:deploymentId', component: DeploymentComponent, title: 'Deployment - Liima' },
   {
     path: 'deployment/:appserverName/:releaseName',
     component: DeploymentComponent,
+    title: 'Deployment - Liima',
   },
 ];
--- a/AMW_angular/io/src/app/deployment/deployment-routes.ts
+++ b/AMW_angular/io/src/app/deployment/deployment-routes.ts
@@ -1,11 +1,12 @@
 import { DeploymentComponent } from './deployment.component';
 
+const DEPLOYMENT_TITLE = 'Deployment - Liima';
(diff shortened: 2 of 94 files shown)


BeriBoss:
Hi Yves.
Thank you for your submission. We will take a look at your PR and address the points listed as soon as possible.

BeriBoss:
Hi Yves

I have a question on this point of yours:
`Searching for an appServer/App doesn't reload the pagination and doesn't show a result when clicked.`

I have had a closer look at this: The frontend correctly implements pagination. The backend endpoint responsible for retrieving all app servers with the applied filters is not introducing any new logic; it simply calls an existing business logic method that has remained unchanged since the migration.

This method creates a query that includes pagination parameters, meaning the pagination is handled directly in the query. However, in this specific case, this approach does not work correctly.

We have two options:

1. Leave it as is, since this issue was probably already present in the previous version of Liima.
2. Merge my fix on branch [bugfix/1077-apps-page-reload-pagination](https://github.com/liimaorg/liima/pull/840) (needs to be reviewed). In this fix, pagination is still handled in the backend, but outside the query. The downside is that the initial page load takes longer.

Would you be okay with this tradeoff, so we can proceed with reviewing and merging the fix?

yvespp:
@BeriBoss can this be merged now?

llorentelemmc:
@yvespp We have decided it would be good to have one final code review of this large PR. After this, we will merge. Hope this is okay for you?

yvespp:
@llorentelemmc sure. As I'm the creator of this pr, technically you have to approve it. Please approve the PR and merge it yourself after the review is done.

AMW_angular/io/src/app/auditview/auditview.routes.ts
```diff
@@ -1,6 +1,6 @@
 import { AuditviewComponent } from './auditview.component';
 
 export const auditviewRoutes = [
-  { path: 'auditview', component: AuditviewComponent },
-  { path: 'auditview/:resourceId', component: AuditviewComponent },
+  { path: 'auditview', component: AuditviewComponent, title: 'Audit View - Liima' },
```
StephGit, line 4: Cosmetics: Use a Constant for both

AMW_angular/io/src/app/deployment/deployment-routes.ts
```diff
@@ -1,10 +1,11 @@
 import { DeploymentComponent } from './deployment.component';
 
 export const deploymentRoutes = [
-  { path: 'deployment', component: DeploymentComponent },
-  { path: 'deployment/:deploymentId', component: DeploymentComponent },
+  { path: 'deployment', component: DeploymentComponent, title: 'Deployment - Liima' },
```
StephGit, line 4: Cosmetics: Use a Constant for both

AMW_angular/io/src/app/resources/resources-list/resources-list.component.html
StephGit, line 1: Maybe styling for app-table column width. Now column sizes switch on each selection

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/ResourceTypesRest.java
```diff
@@ -70,6 +71,20 @@ public List<ResourceTypeDTO> getAllResourceTypes() {
                 .collect(Collectors.toList());
     }
 
+    @Path("/resourceTypes/{name}")
+    @GET
+    @ApiOperation(value = "Get a resource type by name")
+    public ResourceTypeDTO getResourceTypeByName(@PathParam("name") String name) throws NotFoundException, ValidationException {
```
StephGit, line 77: Add @NotNull and remove exception handling for empty value.
Use UseCase and throw exception in service.

StephGit (commented):
Great work throughout. 

Some minor inputs.

Please run `prettier`, there are some formatting issues

llorentelemmc (approved):

PR #774, opened 2024-09-25
Feature/769 angular functions

llorentelemmc:

PR #828, opened 2025-01-15
Feat/1055 edit resource add edit template

BeriBoss:


AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/ResourceTemplatesRest.java
```diff
@@ -122,12 +131,84 @@ public Response getResourceTypeTemplatesById(@PathParam("resourceTypeId") Intege
         return Response.status(Response.Status.OK).entity(templateDTOs).build();
     }
 
+    @POST
+    @Path("/addForResource/{id : \\d+}")
+    @ApiOperation(value = "Add new resource template")
+    public Response addNewResourceTemplate(@ApiParam("Resource ID") @PathParam("id") Integer id, TemplateDTO request)
+            throws AMWException {
```
llorentelemmc, line 138: Best not to pass AMWExceptions to the REST Endpoint: catch AMWExceptions and throw a more specific exception. This way, the user has more information on the reason why a request hasn't worked.
Similar issue: https://github.com/liimaorg/liima/pull/774#discussion_r1786069812
Solution: https://github.com/liimaorg/liima/pull/774/files#diff-ac5c068aa9ee34415e8b771f864a453999ba726eda3ef89e480f89aaa7305ce5L106

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/ResourceTemplatesRest.java
```diff
@@ -122,12 +131,84 @@ public Response getResourceTypeTemplatesById(@PathParam("resourceTypeId") Intege
...
+    public Response modifyResourceTypeTemplate(@ApiParam("Resource ID") @PathParam("id") Integer id, TemplateDTO request) throws AMWException, OptimisticLockException {
+        TemplateDescriptorEntity template = toTemplateDescriptorEntity(request, null);
+        templateEditor.saveTemplateForResourceType(template, id);
+        return Response.created(URI.create("resources/template/updateForResource" + template.getId())).build();
+    }
+
+    @GET
+    @Path("/targetPlatforms/{contextId: \\d+}")
+    @ApiOperation(value = "Get all targetPlatforms for a context id")
+    public Response getTargetPlatformsForContextId(@PathParam("contextId") Integer contextId) {
```
llorentelemmc, line 175: throw NotFoundException if no targetPlatforms found?

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/ResourceTemplatesRest.java
BeriBoss: I think I understand what you're saying, and that's actually what I was wondering - if there's anything else I need to do here. 
The thing is: I've tested it and it works:

image

I have seen in the code that ValidationExceptions are thrown with specific error messages:

Screenshot from 2025-01-23 18-39-30

But an AMWException can also be thrown. And since ValidationExceptions are also AMWExceptions, and AMWExceptions are more general, you can just leave the AMWException in the throws list and don't have to list both. (As you can see in the screenshot, this is what Intellij suggests.) But then I wondered if the mapper would still catch a ValidationException, which is a more specific AMWException. As you can see in the picture above, I would say that this is exactly the case. But please correct me if I have  somehow missed something.

So I would say in this case there is no need to change anything? What do you think?

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/ResourceTemplatesRest.java
BeriBoss: No, you're right, thank you! I will remove it

llorentelemmc (approved):

PR #840, opened 2025-02-17
Bugfix/1077 apps page reload pagination

BeriBoss:
**Bug**: Incorrect Pagination on the Apps Page

The pagination on the Apps page was faulty. When on the page, a list of app servers with their apps was displayed, but neither the correct number of app servers (not the selected maximum results per page) nor the correct number of pages for all results was displayed.

**Incorrect Behavior**: If the maximum number of results per page was set to 10 and there were a total of 50 results, 7 app servers were displayed on the first page, but it was indicated that there were 5 pages. On the next page, 3 app servers were displayed, etc.

**Correct Solution**: 10 app servers should be displayed per page and it should be indicated that there are 5 pages.

**Cause of the Error**: The error was in the backend. The idea was that the frontend could make a REST call and specify the following query parameters:

- start
- limit
- appServerName
- releaseId

The endpoint should then return the filtered result with the correct offset (which page am I on) and the correct maximum number of results per page.

**First Problem**: The total number of results ( `result.getB()` ) was always returned as the number of all unfiltered Resource Entities with the ResourceType = APPLICATIONSERVER. 

`ch.mobi.itc.mobiliar.rest.apps.AppsRest#getApps`:

`return Response.status(OK).entity(appServersToResponse(result.getA())).header("X-total-count", result.getB()).build();`

Solution to that would have been:

`return Response.status(OK).entity(appServersToResponse(result.getA())).header("X-total-count", result.getA().size()).build();`

**Second Problem**: Even if the total number of results had been corrected, the pagination would still not have worked. The reason for this was that the endpoint set a query that retrieved all Resource Entities with the ResourceType = APPLICATIONSERVER and the filtered name. The offset and limit were then set in the query:

```java
   if (startIndex != null) {
            query.setFirstResult(startIndex);
        }
        
    if (maxResult != null) {
            query.setMaxResults(maxResult);
        }
```

The idea was to reduce the load time by only retrieving the "paginated" results directly. However, the problem was that this already paginated list of results was filtered twice in the business logic, so that the set _start_ and _limit_ no longer matched the final list.

**Solution**: The original solution would have been to correct the pagination, but this would have increased the load time (See commit: b695fa063722f5a6e42141a63c82bb0c3e9aca06 ). Since the DB was already under load, it was decided to **remove the pagination and initially display only the first 20 app servers**, as it was in the old version.

### DIFF
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/domain/applist/ApplistScreenDomainServiceQueries.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/domain/applist/ApplistScreenDomainServiceQueries.java
@@ -33,7 +33,6 @@ import ch.puzzle.itc.mobiliar.business.resourcerelation.entity.ConsumedResourceR
 import ch.puzzle.itc.mobiliar.business.utils.JpaWildcardConverter;
 import ch.puzzle.itc.mobiliar.business.utils.database.DatabaseUtil;
 import ch.puzzle.itc.mobiliar.common.util.DefaultResourceTypeDefinition;
-import ch.puzzle.itc.mobiliar.common.util.Tuple;
 
 public class ApplistScreenDomainServiceQueries {
 
--- a/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/domain/applist/ApplistScreenDomainServiceQueries.java
+++ b/AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/domain/applist/ApplistScreenDomainServiceQueries.java
@@ -43,16 +43,10 @@ public class ApplistScreenDomainServiceQueries {
     @Inject
     private DatabaseUtil dbUtil;
 
-
-    Tuple<List<ResourceEntity>, Long> getAppServersWithApps(Integer startIndex, Integer maxResult, String nameFilter) {
+    Tuple<List<ResourceEntity>, Long> getAppServersWithApps(String nameFilter) {
         CriteriaBuilder cb = entityManager.getCriteriaBuilder();
         Predicate p;
         boolean nameFilterIsEmpty = nameFilter == null || nameFilter.trim().isEmpty();
-
-
-
-        // Count all values before filtering
-        Long totalCount = getTotalCount(cb);
 
         // Filter and retrieve results
         CriteriaQuery<ResourceEntity> q = cb.createQuery(ResourceEntity.class);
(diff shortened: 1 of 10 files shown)


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/resourcegroup/boundary/ResourceRelations.java
```diff
@@ -50,11 +50,9 @@ public class ResourceRelations {
     ResourceDependencyResolverService dependencyResolverService;
 
     @HasPermission(permission = Permission.APP_AND_APPSERVER_LIST)
-    public Tuple<List<ResourceWithRelations>, Long> getAppServersWithApplications(Integer startIndex, Integer maxResults, String filter, ReleaseEntity release) {
-        Tuple<List<ResourceEntity>, Long> result = applistScreenDomainService.getAppServerResourcesWithApplications(startIndex, maxResults, filter, true);
-        List<ResourceEntity> appServersWithAllApplications = result.getA();
-        List<ResourceWithRelations> filteredResult = filterAppServersByRelease(release, appServersWithAllApplications);
-        return new Tuple<>(filteredResult, result.getB());
+    public List<ResourceWithRelations> getAppServersWithApplications(String filter, ReleaseEntity release) {
+        List<ResourceEntity> result = applistScreenDomainService.getAppServerResourcesWithApplications(filter, true);
```
llorentelemmc, line 54: Currently the withAppServerContainer boolean is only ever set to true, meaning the filtering logic in AppListScreenDomainService.filterAppServersResourcesWithApplications has no effect

AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/domain/applist/ApplistScreenDomainServiceQueries.java
```diff
@@ -85,27 +79,17 @@ Tuple<List<ResourceEntity>, Long> getAppServersWithApps(Integer startIndex, Inte
...
         TypedQuery<ResourceEntity> query = entityManager.createQuery(q);
 
-        if (startIndex != null) {
-            query.setFirstResult(startIndex);
-        }
+        /*
+         * The maximum number of results to return when no name filter is applied.
+         * This is used to prevent large result sets from being returned.
+         */
+        int DEFAULT_MAX_RESULTS_OF_NON_FILTERED_RESULT = 20;
```
llorentelemmc, line 86: Right now, when no filter is applied the frontend is showing 6 instead of 20 app servers, because some app servers have a large number of applications. Something else is limiting the max results on the page. Should we check what this is?

BeriBoss: I've checked this, and it's related to the 'second problem' I mentioned earlier in the description. The idea is to initially show only a limited number of results so that the user will need to use the filter function anyway -  according to the requirement:

> ...kann man auch einfach zum Start ein Dummy resultat anzeigen.

> ...in dem Fall einfach die ersten 20 in der Tabelle, der User wird damit nicht viel anfangen können.

Therefore, I think the focus should be on limiting the number of results retrieved from the database, rather than limiting the list of displayed results. Otherwise, we’d end up with the same problem as before, wouldn’t we?

What do you suggest we do? @llorentelemmc

llorentelemmc (approved):

PR #841, opened 2025-02-18
feat: improve ui, make app table more consistent with the others

BeriBoss:
This PR should resolve the following 2 issues:

`Apps page: Somehow the table has a lighter gray for the table striping than the other pages. Also the table header is colored but not on the other pages.`

`Apps page: Can you please decrease the margines between the apps and the appServers in the table, so it gets more compact.`

I would appreciate it if this could be checked in the review.


llorentelemmc (approved):

llorentelemmc:
Apps table styling looks consistent with the others!

PR #844, opened 2025-03-04
fix: appServer-dropdown when creating apps

BeriBoss:
**Problem fixed by this PR**: When a new AppServer is created on the page, it cannot be selected from the AppServer-dropdown when creating a new app without reloading the page.


AMW_angular/io/src/app/apps/apps.component.ts
```diff
@@ -85,6 +80,9 @@ export class AppsComponent implements OnInit, OnDestroy {
   lastPage = computed(() => Math.ceil(this.appsService.count() / this.appsService.limit()));
 
   constructor() {
+    this.appServerResourceType$.subscribe((asResourceType) => {
```
llorentelemmc, line 83: Is this necessary in constructor and ngOnInit? I just removed it in constructor and the fix still works:
```
this.appServerResourceType$.subscribe((asResourceType) => {
      this.resourceService.setTypeForResourceGroupList(asResourceType);
    });
```

AMW_angular/io/src/app/apps/apps.component.ts
```diff
@@ -94,6 +92,10 @@ export class AppsComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit(): void {
+    this.appServerResourceType$.subscribe((asResourceType) => {
```
llorentelemmc, line 95: unsubscribe with `.pipe(takeUntil(this.destroy$))`

llorentelemmc (approved):

PR #944, opened 2026-01-20
ResourceEdit: Context-List 

StephGit:
Adds context-list for resource-/type-edit-page

### DIFF
--- a/AMW_angular/io/src/app/deployment/environment.service.ts
+++ b/AMW_angular/io/src/app/deployment/environment.service.ts
@@ -59,14 +59,8 @@ export class EnvironmentService extends BaseService {
       .filter((environment) => environment.parentName === parentName)
       .map((environment) => {
         return {
-          id: environment.id,
-          name: environment.name,
-          nameAlias: environment.nameAlias,
-          parentName: environment.parentName,
-          parentId: environment.parentId,
+          ...environment,
           children: this.buildEnvironmentTree(environments, environment.name),
-          selected: environment.selected,
-          disabled: environment.disabled,
         };
       });
   }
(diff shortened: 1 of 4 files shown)


AMW_angular/io/src/app/deployment/environment.service.ts
```diff
@@ -53,6 +54,23 @@ export class EnvironmentService extends BaseService {
...
+          id: environment.id,
+          name: environment.name,
+          nameAlias: environment.nameAlias,
+          parentName: environment.parentName,
+          parentId: environment.parentId,
+          children: this.buildEnvironmentTree(environments, environment.name),
+          selected: environment.selected,
+          disabled: environment.disabled,
+        };
+      });
```
mburri: I'd write this like:

```
return {
  ...environment,
  children: this.buildEnvironmentTree(environments, environment.name),
};
```
but that is a personal preference - you're free to keep it. And maybe I missed something and my implementation would not work properly.

AMW_angular/io/src/app/resources/contexts-list/contexts-list.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  permissions = computed(() => {
+    if (this.authService.restrictions().length > 0) {
+      return {
+        canView: this.authService.hasPermission('ENV_PANEL_LIST', 'ALL'),
+      };
+    } else {
+      return {
+        canView: false,
+      };
+    }
```
mburri: the else could be dropped. Just do a 
```
return {
    canView: false,
}
```
in the end.

mburri (changes requested):
See my comments on some parts of the code.

PR #964, opened 2026-03-03
Feat/resource properties

StephGit:
This PR adds a new "Properties"-Tile to the edit resource page.

new:

old:

What is different to the old implemenation:

 - Properties are displayed in two columns to provider a better ui
 - Reset checkbox is replaced by a button
 - Color code for upper context is blue instead of green. ('green' normaly states success)
 - Saving is disabled as long as invalid input fields exist
 - editable properties (defined on instance) have actions to edit and delete

### DIFF
--- a/AMW_angular/io/src/app/resources/properties-list/properties-list.component.ts
+++ b/AMW_angular/io/src/app/resources/properties-list/properties-list.component.ts
@@ -1,7 +1,11 @@
 import { Component, computed, input, output } from '@angular/core';
 import { Property } from '../models/property';
 import { PropertyFieldComponent } from '../property-field/property-field.component';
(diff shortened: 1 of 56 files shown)


AMW_angular/io/src/app/resources/properties-list/properties-list.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    const props = this.properties();
+    const mid = Math.ceil(props.length / 2);
+    return props.slice(0, mid);
+  });
+
+  rightProperties = computed(() => {
+    const props = this.properties();
+    const mid = Math.ceil(props.length / 2);
+    return props.slice(mid);
+  });
```
mburri: This looks suspicious!

Couldn't we just use some kind of grid in the template to control the layout?

mburri (approved):

PR #984, opened 2026-03-30
feat: show resource tags in dedicated tile

mburri:
Moved the more -> tag current state action from the "menu" into its dedicated tile. The existing tags for the resource are shown in this tile too.

The analyze -> test generation action should be moved too, but this will be done in different PR


AMW_business/src/main/java/ch/puzzle/itc/mobiliar/business/configurationtag/control/TagConfigurationService.java
```diff
@@ -37,12 +38,13 @@
...
 import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Root;
+import javax.validation.constraints.NotNull;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 
 @Stateless
-public class TagConfigurationService implements CreateTagUseCase {
+public class TagConfigurationService implements CreateResourceTagUseCase, ListResourceTagsUseCase {
```
StephGit: Only a true Java-JediLegacyNightMaster makes use of UseCase/Command Pattern. Well done you wise old man with a very mature and young soul.

StephGit (approved):

PR #985, opened 2026-03-30
feat: show dependencies of resource

mburri:
This PR adds the functionality to show dependencies of a resource.

notable points are:
- show dependencies is always shown - there is no restriction on resource type or permissions.
- there were no permission checks whatsoever to show dependencies (in the backend). It's available for all authenticated users.
- the table contains an action button to edit the dependency. The context parameter is hard coded to  `ctx=1` - this was already the case for the old implmentation

I decided to rework the page layout after implementing the navigation items - and clean it up over all pages. Now all pages are different.


StephGit (changes requested):
feature is missing for `resource-type.edit`

yvespp (approved):

StephGit (approved):

PR #988, opened 2026-04-01
Fix/property desc issues

StephGit:
fixes the issues from https://github.com/liimaorg/liima/pull/980


AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/PropertyDescriptorRest.java
```diff
@@ -200,46 +138,22 @@ public Response deletePropertyDescriptor(
...
-                    .build();
-        } catch (AMWException e) {
-            return Response.status(Response.Status.BAD_REQUEST)
-                    .entity(e.getMessage())
-                    .build();
-        }
+        return Response.status(Response.Status.NO_CONTENT).build();
     }
 
     private String convertTagsToString(java.util.List<ch.mobi.itc.mobiliar.rest.dtos.PropertyTagDTO> tags) {
```
mburri: use a qualified import for java.util.List and also for the PropertyTagDTO

AMW_rest/src/main/java/ch/mobi/itc/mobiliar/rest/resources/PropertyDescriptorRest.java
```diff
@@ -134,62 +96,38 @@ public Response updatePropertyDescriptor(
...
+        existingEntity.setPropertyName(dto.getName());
+        existingEntity.setDisplayName(dto.getDisplayName());
+        existingEntity.setValidationLogic(dto.getValidationRegex());
+        existingEntity.setNullable(dto.isNullable());
+        existingEntity.setOptional(dto.isOptional());
+        existingEntity.setEncrypt(dto.isEncrypted());
+        existingEntity.setMachineInterpretationKey(dto.getMik());
+        existingEntity.setDefaultValue(dto.getDefaultValue());
+        existingEntity.setExampleValue(dto.getExampleValue());
+        existingEntity.setPropertyComment(dto.getComment());
```
mburri: I don't think that db entities should leak into the controller...

mburri (commented):
Looks better now but I still think that there is way too much logic in the controllers...

IMHO, The rest endpoints should take the input. Transform it into a command (the command does some input validation) and hands it then to a service or use case the is implemented in AMW_business. 

There, exceptions are thrown, e.g. if a resource is not found or some other stuff happens.

I also think that we should move the command into AMW_business. And the service method should just get the command. If you want to do this, you cannot use the DTO in the command - because you cannot (or should not) use stuff from AMW_Rest (the outer ring) in AMW_bussines... (the inner ring)... because depdendencies should only point to inner rings...

mburri (approved):

PR #993, opened 2026-04-21
feat: show, add and remove applications to resources

mburri:


AMW_angular/io/src/app/resources/resource-edit/resource-applications/resource-applications.component.html
```diff
@@ -0,0 +1,109 @@
+<app-loading-indicator [isLoading]="isLoading()"></app-loading-indicator>
+<app-tile-component
+  [title]="'Applications'"
+  [actionName]="'Add Application'"
+  [canAction]="permissions().canUpdate"
+  [isVisible]="permissions().canListRelations"
```
StephGit: this property is used to initiallly show the tilebody open or closed. using a permission is maybe the wrong way. Default is false only overwrite it to enforce open tile body
Should be changed aswell in resource-tags.component.html#L6

the permissions could be used for the signal-input of `notAllowed`, which will display a according message.

AMW_angular/io/src/app/resources/resource-edit/resource-applications/resource-applications.component.html
```diff
@@ -0,0 +1,109 @@
...
+  [noContent]="false"
+  [notAllowed]="false"
+  (tileAction)="showAddApplicationModal()"
+>
+  @if (error()) {
+    <div class="alert alert-danger" role="alert">
+      {{ error() }}
+    </div>
+  } @else if (!applications() || applications().length === 0) {
```
StephGit: This check can be passed to the input signal of `noContent`

Same in `resource-release.component.html#L11` and `resource-tags.component.html#L11` is a check for content.

AMW_angular/io/src/app/resources/resource-edit/resource-applications/resource-applications.component.html
```diff
@@ -0,0 +1,109 @@
...
+        }
+      </tbody>
+    </table>
+  }
+</app-tile-component>
+
+<ng-template #addApplicationModal let-modal>
+  <app-modal-header [title]="'Add Application'" (cancel)="modal.dismiss('Cross click')"></app-modal-header>
+  <div class="modal-body">
+    @if (availableApplications().length === 0) {
```
StephGit, line 61: Would it make more sense to not display the `addApplicationButton` if there are not apps available, instead of checking this in the modal?

AMW_angular/io/src/app/resources/resource-edit/resource-applications/resource-applications.component.ts
```diff
@@ -0,0 +1,163 @@
...
+
+      this.loadApplications(id);
+    });
+  }
+
+  private loadApplications(resourceId: number): void {
+    this.isLoading.set(true);
+    this.error.set(null);
+
+    this.resourceApplicationsService.getApplicationsForResource(resourceId).subscribe({
```
StephGit: why not using signals instead of observables throughout the component?

With all new components there should be signals prefered over observables. And if not, the cleanup pattern with takeUntil(this.destroy$) is missing:

- CopyFromResourceDialogComponent
- ResourceReleasesComponent
- ResourceTagsComponent

StephGit (changes requested):

AMW_angular/io/src/app/resources/resource-edit/resource-applications/resource-applications.component.ts
mburri: For now, I just change it for resource applications

StephGit (approved):

PR #1004, opened 2026-06-03
fix accessibilty issues

StephGit:


AMW_angular/io/src/app/resources/resource-edit/resource-releases/resource-releases.component.html
```diff
@@ -7,54 +7,25 @@
...
+    [canNavigate]="false"
+    [canDelete]="permissions().canDeleteRelease"
+    [canEdit]="permissions().canChangeRelease"
+    [rowStyle]="rowStyle()"
+    [canRowClick]="true"
+    (rowClick)="onRowClick($event)"
+    (navigate)="onTableAction($event)"
+    (edit)="onTableAction($event)"
+    (delete)="onTableAction($event)"
+  ></app-table>
```
mburri, line 23: While I agree that the previouse version might have some problems, a component with 12 input props is  also not easy to understand...

AMW_angular/io/src/app/shared/tile/tile-list/tile-list.component.ts
```diff
@@ -21,7 +21,7 @@ export interface TileListEntryOutput {
 @Component({
   selector: 'app-tile-list',
   template: `
-    <div class="title">{{ title() }}</div>
+    <h3 class="title">{{ title() }}</h3>
```
mburri, line 24: Could thy styling now be done using the h3 tag itself instead of using a `title` class?

mburri (approved):
I think the changes to the release table (use the table component) are not really a11y improvements - are they?
It would have been nice to have them in two separate PRs - now it's mixed up...

PR #1005, opened 2026-06-08
Feat/add remove relation (974)

StephGit:

PR #1006, opened 2026-06-09
Feat/related applications (1202)

StephGit:
Show active applications in related-resources tile for APPLICATIONSERVERS on Node relation.
Example:
http://localhost:8080/AMW_web/pages/editResourceView.xhtml?rel=360266&ctx=1&id=263978

Needs merge of https://github.com/liimaorg/liima/pull/1005 first


mburri (changes requested):
I get an error when selecting the node... and there are no active applications...

Image

mburri (approved):

PR #1009, opened 2026-06-10
fix: back navigation to resource page (1067)

StephGit:


AMW_angular/io/src/app/resources/resources-page.component.ts
```diff
@@ -58,25 +58,30 @@ export class ResourcesPageComponent implements OnDestroy, OnInit {
...
+      const idFromUrl = this.selectedResourceTypeIdFromUrl();
+      if (idFromUrl === null) return;
+
+      const allTypes = [...(this.predefinedResourceTypes() || []), ...(this.rootResourceTypes() || [])];
+      if (allTypes.length === 0) return;
+
+      const resourceType = allTypes.find((rt) => rt.id === idFromUrl);
+      if (resourceType) {
+        this.toggleChildrenAndOrLoadResourcesList(resourceType, false);
+      }
```
mburri: using `effect` to sinchronize app state is an anti pattern.

we could get the `selectedResourceTypeIdFromUrl` like this:

```
private selectedResourceTypeIdFromUrl = toSignal(
    this.route.queryParamMap.pipe(
      map(params => {
        const id = params.get('selectedResourceTypeId');
        return id ? Number(id) : null;
      })
    ),
    { initialValue: null }
  );
```

and then move the logic to select the resourceType and load the resource list out of the constructor/ effect

```
// 2. Reactively find and auto-select the resource type when EITHER the URL or Data changes
  private autoSelectTrigger = computed(() => {
    const idFromUrl = this.selectedResourceTypeIdFromUrl();
    if (idFromUrl === null) return;

    const allTypes = [...(this.predefinedResourceTypes() || []), ...(this.rootResourceTypes() || [])];
    if (allTypes.length === 0) return;

    const resourceType = allTypes.find((rt) => rt.id === idFromUrl);
    if (resourceType) {
      // Safely execute selection out-of-stack to avoid expression-changed bugs
      setTimeout(() => this.toggleChildrenAndOrLoadResourcesList(resourceType, false));
    }
  });

  constructor() {
    // Simply reference the computed signal so it stays active and listens for changes
    this.autoSelectTrigger();
  }
```

This also renders the `$onDestroy` observable obsolete...

+ some more refactorings might then be requiered

mburri (changes requested):

mburri (approved):

PR #1010, opened 2026-06-10
fix: autoload filters and improve filter mapping for deployments

### LINKED ISSUE #997: Some Deployment filters don't work anymore
Opening this URL doesn't set the filter from the URL anymore and the filter gets deleted in the URL:

`https://<host>/AMW_angular/#/deployments?filters=[{"name":"Tracking Id","comp":"eq","val":"2601987"}]`

Number based filters like Tracking Id, Id seem to be broken affect by this.

Date filters don't filter correctly anymore for example this shows only a deployment with scheduled date of 2036:

`https://<host>/AMW_angular/#/deployments?filters=[{"name":"Deployment date","comp":"gte","val":"14.05.2026 22:45"}]`

The Reason filter will just be removed if you click `Apply filter`.

StephGit:
Fixes https://github.com/liimaorg/liima/issues/997

### DIFF
--- a/AMW_angular/io/src/app/deployments/deployments.component.ts
+++ b/AMW_angular/io/src/app/deployments/deployments.component.ts
@@ -57,11 +57,15 @@ export class DeploymentsComponent implements OnInit {
   filterTypes = signal<DeploymentFilterType[]>([]);
   private filterTypeMap = computed(() => {
     const map = new Map<string, FilterType>();
+    const validFilterTypes = Object.values(FilterType);
     this.filterTypes().forEach((filterType) => {
-      const validFilterTypes = Object.values(FilterType);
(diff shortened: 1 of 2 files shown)


AMW_angular/io/src/app/deployments/deployments.component.ts
```diff
@@ -487,12 +490,12 @@ export class DeploymentsComponent implements OnInit {
           this.parseDateTime(filter, filterType);
           enhancedFilters.push(filter);
         } else {
-          this.errorMessage = 'Error parsing filter';
+          this.errorMessage = `Unknown filter: "${filter.name}"`;
```
mburri, line 493: 🌈

i really like good error messages

mburri (approved):

PR #1017, opened 2026-06-22
Update to Angular 22

yvespp:
- Resolve all TypeScript strict errors after Angular upgrade
- Replace deprecated baseUrl with paths mapping
- Move side effects out of computed() into effect()
- Replace template non-null assertions with proper type narrowing
- Replace silent null masking with explicit error handling
- Update JS deps

With ts 6 there are a lot of errors thrown in the project.
I tried fixing them using AI but I'm not sure if they are ok.


AMW_angular/io/src/app/apps/app-add/app-add.component.ts
```diff
@@ -17,13 +17,13 @@ import { ButtonComponent } from '../../shared/button/button.component';
 export class AppAddComponent {
   activeModal = inject(NgbActiveModal);
 
-  @Input() releases: Signal<Release[]>;
-  @Input() appServerGroups: Signal<Resource[]>;
+  @Input() releases!: Signal<Release[]>;
```
StephGit: We should prefer `input.required` in those cases.

```suggestion
releases = input.required<Release[]>();
```

AMW_angular/io/src/app/apps/app-add/app-add.component.ts
```diff
@@ -17,13 +17,13 @@ import { ButtonComponent } from '../../shared/button/button.component';
...
+  @Input() appServerGroups!: Signal<Resource[]>;
```
StephGit: we should use `input.required` in those cases

AMW_angular/io/src/app/apps/app-add/app-add.component.ts
```diff
@@ -17,13 +17,13 @@ import { ButtonComponent } from '../../shared/button/button.component';
...
   @Output() saveApp: EventEmitter<AppCreate> = new EventEmitter<AppCreate>();
 
-  app: AppCreate = { appName: '', appReleaseId: null, appServerId: null, appServerReleaseId: null };
-  appServerGroup: Resource;
-  appServerRelease: Rel;
+  app: AppCreate = { appName: '', appReleaseId: null as unknown as number, appServerId: null as unknown as number, appServerReleaseId: null as unknown as number };
```
StephGit: could be null as defined in interface. this is a code-smell

AMW_angular/io/src/app/apps/app-add/app-add.component.ts
```diff
@@ -17,13 +17,13 @@ import { ButtonComponent } from '../../shared/button/button.component';
...
+  appServerGroup!: Resource;
```
StephGit: should be `Resource | undefined = undefined;`

StephGit (approved):

PR #1031, opened 2026-07-29
fix: reset form when canceling edit template

BeriBoss:


AMW_angular/io/src/app/resources/resource-edit/resource-templates/resource-template-edit/resource-template-edit.component.ts
```diff
@@ -49,9 +49,11 @@ export class ResourceTemplateEditComponent {
 
   private _template!: ResourceTemplate;
```
StephGit: might be less confusing (at least for me) if we could use more clear names like old/new than _template/template

StephGit (approved):

AMW_angular/io/src/app/resources/resource-edit/resource-templates/resource-template-edit/resource-template-edit.component.ts
BeriBoss: I renamed it to `_currentWorkingTemplate` to make it clearer that this represents the template being actively edited in the component.

PR #1033, opened 2026-07-29
Fix/selected resource typte id url param

BeriBoss:
# Fix URL parameter selection bug and refactor resource selection logic

- Fixed URL parameter `selectedResourceTypeId` being ignored on resources page by changing `autoSelectTrigger` from computed to effect
- Added default selection to `autoSelectTrigger` to select first root type when no URL parameter is provided
- Removed `selectedResourceTypeOrDefault` and redundant `selection` property to eliminate race conditions
- Code quality improvements: renamed methods, simplified array checks, fixed error messages
- Updated template to use `selectedResourceType()` signal consistently

### DIFF
--- a/AMW_angular/io/src/app/resources/resources-page.component.html
+++ b/AMW_angular/io/src/app/resources/resources-page.component.html
@@ -10,7 +10,7 @@
             <li class="nav-item">
               <a
                 class="nav-link ps-0"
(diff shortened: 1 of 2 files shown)


StephGit (changes requested):
The fix adds a new bug, when selecting a parent-type with children, users have to click twice before list is collapsed

PR #1034, opened 2026-08-03
fix: restrict relation types for non-default resources

BeriBoss:
Fixes unresolved related resources logic for non-default resource types. Non-default resource types can now only add relations to resource types defined at the type level (via unresolved relations). Default resource types (`APPLICATION`, `APPLICATIONSERVER`, `NODE`, `RUNTIME`) continue to allow adding any resource type.

**Note**: The JSF version hides all related resources for `NODE` resource types entirely (`rendered="${!resourceRelationModel.node}"`), and this PR now fixes this issue by adding an `isNode()` check to hide the relations component for `NODE` resource types.

### DIFF
--- a/AMW_angular/io/src/app/resources/resource-edit/resource-relations/resource-relations.component.ts
+++ b/AMW_angular/io/src/app/resources/resource-edit/resource-relations/resource-relations.component.ts
@@ -197,6 +197,12 @@ export class ResourceRelationsComponent extends BaseRelationsDirective {
     return isAppServer && isNode;
   });
 
+  protected isNode = computed(() => {
+    const type = this.resource()?.type;
+    const cleanType = type?.replace(/"/g, '');
+    return cleanType === 'NODE';
+  });
+
   protected properties = computed<Property[]>(() => {
     const props = this.relationsService.relationProperties;
     const result: Property[] = [];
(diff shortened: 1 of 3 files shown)


AMW_angular/io/src/app/resources/resource-edit/resource-relations/resource-relations.component.ts
```diff
@@ -199,6 +197,12 @@ export class ResourceRelationsComponent extends BaseRelationsDirective {
     return isAppServer && isNode;
   });
 
+  protected isNode = computed(() => {
+    const type = this.resource()?.type;
+    const cleanType = type?.replace(/"/g, '');
```
StephGit, line 202: Could be used as seperate function on all those string checks in this file

AMW_angular/io/src/app/resources/resource-edit/resource-relations/resource-relations.component.ts
```diff
@@ -272,13 +276,13 @@ export class ResourceRelationsComponent extends BaseRelationsDirective {
...
     this.availableResourceGroups.set([]);
-    this.resourceTypesService.getRootResourceTypes().subscribe({
-      next: (types) => this.resourceTypes.set(types),
-      error: (err) => {
-        console.error('Failed to load resource types:', err);
-        this.toastService.error('Failed to load resource types.');
-      },
-    });
+
+    if (this.isDefaultResourceType()) {
```
StephGit, line 280: What does this add for non-default resources? 
Relations are not visible for NODE-types?

StephGit (approved):

